PR #528, opened 2015-03-13
Update and rename README.txt to README.md

starit:


pmaupin:
I decided to keep the readme as restructured text (and rename it rst).  Please double-check and see if you are happy with it now.

Thanks,
Pat

oz123:
Installing should not require sudo. Hence closing.

PR #541, opened 2015-07-06
Add initial travis ci tests

pmaupin:
@aquavitae 

SO, I did the cheesiest possible thing to get a few tests going.  I just created a travis ci matrix with a single entry in it, and selected a few packages that made most things work.

I used anaconda, which seems to work pretty well (pre-installed version of PIL).

The build actually fails, because I'm just doing a compare of a generated log file against one I saved.  They differ in the "unknown" section, because apparently (maybe only for sphinx) some of the checksums wobble around.  I have no idea why yet...

Anyway, 206 passing tests that made identical checksums to the ones @ralsina had saved seems like a good start.

At the end of the day, one of us should probably look into resuscitating his nose tests.  At the moment, I want to do some analysis so we can move forward with a later version of reportlab and the other libs, and then that will allow us to do a cleaner upgrade to Python 3, I think.


pmaupin:
@aquavitae 

I have a pretty good python 2.7 matrix running now -- just dumping PDFs and showing good/bad/unknown/failing checksums.  There are obvious gratuitous output changes in reportlab that need to be checked, but there are a *lot* of running tests just based on the checksums Roberto had entered as good.

You can see all this if you look at the test results:

https://travis-ci.org/rst2pdf/rst2pdf/builds/69839982

There were actually gratuitous output changes caused by the pdfrw 0.1 to 0.2 transition, when I blindly accepted a fix (and didn't have these tests going), but I put some code into the vectorpdf extension module to revert this change.

If you clone the rst2pdf/test_artifacts, you can see where I dumped the files, under the builds directory:

https://github.com/rst2pdf/test_artifacts/tree/master/builds

That repository has a couple of tools in it that pretty much assume they are running right under the rst2pdf/tests directory, so I just go there to clone it.  One of those tools is what I used to run the PDFs through a cleaner (which we can maybe add to) and make the output PDF MD5s match the input PDF MD5s.  The other tool will download the artifacts from the travis CI builds into the builds subdirectory.  I don't think we need to keep all those, but for now it was easy and small enough to put them all in there to see differences.

I still haven't tried Roberto's nose test runner.  It's probably time to hook that up, and figure out what we want to do about slightly different PDFs.

Also, to add other Python versions into the mix for the CI, you probably want to take a look at the URLs I put at the top of the .travis.yml file.  I think it's pretty easy.

Anyway, I've done about all I can do with this stuff (except answer questions) for a few weeks.

Thanks,
Pat

akrabat:
Superseded by #624

Issue #525, opened 2015-03-12
CreationDate metadata shows non-reproducible string representation of object instead of data (using sphinx)


GoogleCodeExporter:
```
The people working on reproducible builds on Debian found out that the 
monkey-patched PDFDate does not extend PDFObject, so Reportlab formats it using 
str instead of .format(), resulting in a string like 
"<rst2pdf.createpdf.PDFDate instance at 0x7fc7cd56b320>" instead of a proper 
date, which gets embedded in the CreationDate PDF metadata.

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=777251

I've checked that this happens when running rst2pdf from sphinx, but not when 
running it directly.

The attached patch comes directly from the debian bugreport.

Please answer the following questions, they are *very* important:

* Version of rst2pdf? 0.93

* Version of reportlab? 3.1.8

* Are you using rst2pdf with Sphinx, and if true, what version? 1.2.3

```

Original issue reported on code.google.com by `elena.va...@gmail.com` on 9 Feb 2015 at 1:20

Attachments:
* [patch.diff](https://storage.googleapis.com/google-code-attachments/rst2pdf/issue-525/comment-0/patch.diff)

valholl:
Would turning this patch into a pull request help?

pmaupin:
Sure, I'd review a pull request.  Haven't gone back to look at all the bugs...

oz123:
fixed in #551

PR #551, opened 2015-08-26
Fix embedding of object addresses instead of dates

valholl:
This is the patch attached to issue #525 

Thanks in advance


oz123:
I have reviewed this patch, it seems to work. Test are missconfigured with travis. This helps creates reproducible md5sum of the pdfs.

PR #574, opened 2016-06-19
Use relative path above others

agjohnson:
The behavior being changed here is that images will reference relative images, but
when the `MyImage.size_for_node` method is called, it references the
`self.client.basedir`, which is not going to be the same as `os.getcwd()` given
a path structure like:

```
images/background.pdf
files/test.rst
```

This works for my use case, which uses a path similar to the above example. Any guidance on other side effects this change might induce would be helpful. I ran tests, which were already failing, but I've introduced no new bugs with this PR at least.

### DIFF
--- a/rst2pdf/createpdf.py
+++ b/rst2pdf/createpdf.py
@@ -999,6 +999,9 @@ class FancyPage(PageTemplate):
         info = self.image_cache.get(uri)
         if info is None:
             fname, _, _ = MyImage.split_uri(uri)
+            rel_fname = os.path.join(os.getcwd(), fname)
+            if os.path.exists(rel_fname):
+                fname = rel_fname
             if not os.path.exists(fname):
                 del self.template[which]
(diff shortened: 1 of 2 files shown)


akrabat:
To clarify what this PR does;  it changes the search for images algorithm to look for images in paths relative to the current working directory in preference to paths relative to the directory that the input file is in.

i.e. given `files/test.rst` of:

```rst
.. image:: images/foo.png
```

then:

```
$ rst2pdf files/test.rst -o output.pdf
```

will currently look for `foo.png` in `files/images/` with today's rst2pdf. However, after this PR, it will look in `./images/` first before falling back to  `files/images/` after this patch.

akrabat:
Question: should the order be the other way around?

i.e. should rst2pdf look for a path relative to the input file first and if that fails, then look relative to the current working directory?

akrabat:
#671 replaces this one with rebase to master, tests & other updates…

Issue #537, opened 2015-07-03
python 3 support


flying-sheep:

pmaupin:
So, I haven't really been following along all that well, but at least one person is working on this.  He didn't fork it from Roberto, so it doesn't show up in the graph.  I only know about it because he filed an issue against pdfrw.  Anyway, it's here:

https://github.com/aquavitae/rst2pdf

FWIW, many of the pieces are in place.  reportlab, docutils, pdfrw all support Python 3.  Based on my experience porting pdfrw, I have a few observations:

1) If we limit to 2.6, 2.7, 3.3, 3.4, I don't think it will be all that hard.

2) reportlab keeps changing output and even the setting to ask it not to do that doesn't help.  (To be fair, this may partly be my fault in my pdfrw tests, because I don't think I was ordering dictionaries before wrapping them in reportlab objects).

3) The idea of using MD5 checksums for the tests is still viable, but the right answer may involve round-tripping the reportlab output back through pdfrw and then spitting it back out, because you can probably do that in such a way that the metadata gets stripped and everything else gets placed in a reproducible order.

Thanks,
Pat

flying-sheep:
promising! since this module’s author hasn’t been active for 2 years here, @ralsina should at least add him as collaborator and prepare a release using his work.

ralsina:
@pmaupin has graciously offered to start an organization so we can migrate rst2pdf to a "neutral" place, and then this can move forward.

pmaupin:
Actually, there's not much to creating the organization.  You should already have an invite, Roberto.  Once you accept it, I can either fork the repo and manually (well, there seem to be some scripts available to do it) copy the issues over, or you could transfer ownership of the repo (and then clone it back if you want) so that the issues automatically move over.

Transferring is easy -- I transferred the astor repository to berkerpeksag when I wasn't interested in it.  You just go to the repo, click on the screwdriver/wrench icon  on the lower-right, and then scroll down and press the transfer button.

ralsina:
Done.

flying-sheep:
cool!

so what’s left before merging @aquavitae’s work?

pmaupin:
Well, for one thing, you could just go and use his work now :)

In terms of merging it into master (and then later doing a release), I have had a private conversation with him and I think we agree that the right answer is going to be to treat what he has done as an advance-development branch, and to move a bit more slowly.

[...]

PR #575, opened 2016-07-16
[FIX] Python2-only except clauses

oesteban:
The ```except Exception, val``` syntax is not supported in python 3. A small bite in #537


oz123:
@oesteban rst2pdf is slowly coming back to life from dormancy. Thanks for the pool request. I hope you keep contributing.

oesteban:
Sure thing!

PR #583, opened 2016-08-19
[FIX] Add .travis.yml to execute tests

oesteban:
Python 2.7 tests pass (most of them are skip, I don't know why)
Python 3.5 are, as expected, broken

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+*.pyc
+__pycache__
+*.egg-info
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 *.pyc
 __pycache__
 *.egg-info
+bin/
+develop-eggs/
+eggs/
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ python:
 - 3.5
 
 before_install:
-- pip install nose reportlab Pygments sphinx pillow pdfrw pyPdf Jinja2 coverage
+- pip install nose reportlab Pygments sphinx pillow pdfrw pyPdf Jinja2 coverage wordaxe
 - python setup.py install
 - python bootstrap.py
 - ./bin/buildout
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,17 +4,11 @@ python:
 - 3.5
 
 before_install:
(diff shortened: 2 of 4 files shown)


oz123:
Wow, that is awesome you did this.

The tests are not easy to figure. It took me a while to figure how to run them.

Check the file https://github.com/rst2pdf/rst2pdf/blob/master/doc/DEVELOPERS.txt

One of the long term goals is definitely making this more pythonic by using py.test and setup tools. 

Also, I am not quite sure conda is what I would like to use. I rather use vanilla setup tools and remove buildout. Are you using it now to configure the sources?

And finally why is wordaxe removed?

oz123:
Travis should run the following commands from https://github.com/rst2pdf/rst2pdf/blob/master/doc/DEVELOPERS.txt#L96 onwards.

The test will fail except for those I have already updated. 
Tomorrow I will add a note on how to update the tests such they don't fail.

oesteban:
Hi @oz123 , I have updated the command line for tests in travis. I hope it works now as expected.

Regarding the open questions:
  - I generally prefer to use a conda distribution because most of the necessary software is precompiled and saves tons of time fixing problems (and deployment in travis). Of course it can be done using the vanilla python provided with travis, but having Pillow and those kinds of dependencies I think conda can ease this a lot.
  - Regarding wordaxe: first of all I did not find install candidates in pypi neither for 2.7 nor 3.5. Additionally, I just removed the imports from here: https://github.com/rst2pdf/rst2pdf/pull/583/files#diff-8b36b118be021b3da981b6c41ff4b7e1L7 . It did not make a lot of sense.

oz123:
Wordaxe is in pypi HTTPS://pypi.python.org/pypi/wordaxe/1.0.1
Here is a patch for Python 3
https://sourceforge.net/p/deco-cow/patches/6/
This is related then to #585

oz123:
Ok. I know why your build fails. Travis installs builtout and nose into the virtual environment, hence `nose` is already in your path. No need to call it with `./bin` prefix. If you fix this you'll be 100% signed on the first running CI build 😉. 

Then I will gratefully merge your PR!

oz123:
Thanks you very much!

PR #588, opened 2016-08-24
[PY3] Fixed tests with futurized

oesteban:
~~This is ready to merge.~~


oesteban:
I am looking at the md5 sums of files and they are never the same. This replicates with v0.93. I think we should aim at fix that before merging any new thing.

PR #607, opened 2017-01-16
Fix configparser use for Python2

JoakimSoderberg:
http://python-future.org/compatible_idioms.html#configparser

```python
# Python 2 only:
from ConfigParser import ConfigParser

# Python 2 and 3 (after ``pip install configparser``):
from configparser import ConfigParser
```

Note that, as of v0.16.0, python-future no longer includes an alias for the configparser module because a full backport exists (see https://pypi.python.org/pypi/configparser).

### DIFF
--- a/rst2pdf/config.py
+++ b/rst2pdf/config.py
@@ -8,8 +8,7 @@ from __future__ import absolute_import
 from future import standard_library
 standard_library.install_aliases()
 
-
-import configparser
+from configparser import ConfigParser
 import os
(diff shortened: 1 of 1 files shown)


JoakimSoderberg:
Well actually there will still be an issue with this since Python2 version when installing does not have a dependency for `configparser`.

So the only way to fix this is to have a check for python version, or add the dependency

akrabat:
Merged to dev branch so we don't lose this information.

PR #623, opened 2017-11-03
Support :hl_lines: code directive

akrabat:
As we can't do background colours over whitespace, we dim the lines
that are not highlighted with the new class `pygments_diml`.


akrabat:
Rebased to latest master and added reference PDF as per @alexwlchan's latest tests.

Issue #621, opened 2017-11-02
Reinstate Travis for master


akrabat:
We need to re-instate Travis tests for `master`. At a minimum, It needs to test for Python 2.7 with ReportLap 3.4.0.

For this to happen, every test needs to be run and the output PDF manually checked. If okay, then the hash needs to be updated using `cd tests; python setmd5.py good output/{name-of-test.pdf} `.

akrabat:
Initial work on this here: https://github.com/rst2pdf/rst2pdf/compare/master...akrabat:travis?expand=1

alexwlchan:
I had a quick go at getting Travis running against master here, based on some stuff in the dev branch, but I have no idea how complete a test coverage it might be getting. This was a quick on-the-train-to-work job: https://github.com/alexwlchan/rst2pdf/pull/3

Most of the tests fail with unknown MD5 hashes: https://travis-ci.org/alexwlchan/rst2pdf/builds/297502572

akrabat:
I couldn't get dev to work with python 2, so that's interesting. I got this far with `master`: https://github.com/rst2pdf/rst2pdf/compare/master...akrabat:travis?expand=1

alexwlchan:
> I couldn't get dev to work with python 2, so that's interesting

I didn’t try to get dev running with Python 2, but Travis is running against that branch so I tried to cherry-pick the bits that looked useful from a dev..master diff.

I really don’t know much about the project yet, so this could be totally wrong!

akrabat:
Very strange. I'll have to look at what's going on with my setup!

As per, #618, I think that getting a release out from master is the logical next step though. 

Maybe dev is stable enough to release from though?

akrabat:
Ah, your green-travis branch is from master, which explains why it works :)

akrabat:
In order to minimise overlap of work, I've raised PR #624. The first 15 tests pass and the remainder need md5 hashes updated. Please raise PRs against my `travis` branch to fix things.

lornajane:
If anyone has time to try this on their own machine, I am using this two-line setup to compare my test outcomes with the ones listed above.
[officialfail.txt](https://github.com/rst2pdf/rst2pdf/files/1937254/officialfail.txt) is the list above in a text file.

From `/rst2pdf/tests` try this:

```
nosetests -i regulartest -i sphinxtest 2>&1 | grep "FAIL: " | cut -d' ' -f2 > myfailure.txt
wc -l myfailure.txt
[...]

PR #624, opened 2017-11-05
Enable Travis for master

akrabat:
Addresses #621

Needs all md5s checked and updated before merge.


alexwlchan:
A small number of tests are failing because they can’t find inkscape. Judging from the [Inkscape installation instructions](https://code.launchpad.net/~inkscape.dev/+archive/ubuntu/stable) and [Travis docs](https://docs.travis-ci.com/user/installing-dependencies/), I think adding the following to `.travis.yml` might get those a bit further along:

```yml
before_install:
  - sudo add-apt-repository --yes ppa:inkscape.dev/stable
  - sudo apt-get update
  - sudo apt-get install --yes inkscape
```

alexwlchan:
Also, this error suggests we’re running a newer version of Sphinx than when the tests last passed:

```
* WARNING: sphinx.ext.pngmath has been deprecated. Please use sphinx.ext.imgmath instead.
```

Is it worth trying to get a set of dependencies that are known good, even if they’re quite out-of-date?

PR #626, opened 2017-11-15
First collection of tests with fixed MD5 hashes

alexwlchan:
As well as adding new hashes, I’ve checked in the output PDF files – not for directly comparing in the tests (at least, not yet!) but so the next person to update the MD5s knows what I was looking at.

Related: #621.


alexwlchan:
Previously there were 271 failures. I expect this build to have 255 failures.

alexwlchan:
255 failures, as expected!

akrabat:
Thanks! Tested locally and I also see 255 failures :)

PR #638, opened 2018-01-21
UML test: new checksum but reference PDF is missing

lornajane:
I'm on Ubuntu 17.10 and have the following software versions:
- inkscape: 0.92.2
- plantuml: 8039-1 (what sort of a version number is that?  It came from `apt-get install plantuml`)

The test `test_uml_extension` produces a different md5 than the test suite expects, which isn't a surprise as I am using a newer version of inkscape.  I couldn't find the reference PDF for this test though.  If the attached PDF looks right, then merge this PR to accept the new md5 into the mix and let me know if I should also add the PDF into the repo.  Otherwise, just close or let me know if I can do further testing.
[test_uml_extension.pdf](https://github.com/rst2pdf/rst2pdf/files/1649875/test_uml_extension.pdf)


akrabat:
Installing plantuml via brew, I get the same looking PDF as you do.

Please add the PDF to tests/reference as part of this PR.

PR #639, opened 2018-01-21
Minor additions to testing instructions

lornajane:
Adding a little context of what the tests actually do, and instructions for running one test.

Also: to mark a test as skipped, create `rst2pdf/tests/input/[test].ignore` file (which can be empty).  I couldn't find that documented anywhere and it might be handy!


akrabat:
Thanks.

As part of the merge, I've noted that the ignore file should contain the reason that the test is skipped and have also renamed the file to DEVELOPERS.rst.

PR #640, opened 2018-01-28
pdfbuilder does not return extension metdata for Sphinx

tk0miya:
Since Sphinx-1.3, it expects the extensions to return its metadata on setting up.
http://www.sphinx-doc.org/en/stable/extdev/index.html#extension-metadata

At this moment, Sphinx guesses the metadata of rst2pdf.
But we're planning to drop this special treatment in nearly future (see https://github.com/sphinx-doc/sphinx/pull/4463).

This PR modifies pdfbuilder to follow the protocol for Sphinx extensions.


akrabat:
Thanks @tk0miya

Issue #652, opened 2018-04-24
###Title### gets replaced to empty string


jmones:
I'm using rst2pdf from Sphinx. I'm trying to use `###Title###` in a header by defining defaultHeader attribute in a pageTemplate. While it seems to work for `###Section###` or `###Page###`, `###Title###` is always empty.

I've checked a little bit and `doc.title` in `createpdf.HeaderOrFooter.replace.replace_tokens` seems to be empty.  Also `genelements.HandleTitle.gather_elements` never seems to get into the first if statement (`if isinstance(node.parent, docutils.nodes.document):`), which is the one I assume that should define `client.doc_title_clean` to the correct value.

I've defined the title in the `pdf_documents` definition in the `conf.py` file, which I understood it's the place to define the title of the document.

jmones:
After digging a little bit more, a possible origin of the problem may be that everytime I get into `HandleTitle.gather_elements`, `node.parent` is of type `docutils.node.section` and not `docutils.node.document`. This is also true for the first `rst` file defined in `conf.py` in the `pdf_documents` array.

akrabat:
Using this `.rst` file:

```rst
Test Doc
========


The Section
-----------

The content


.. raw:: pdf

    PageBreak



Another Section
---------------

Some more content
[test_header_footer_in_stylesheet.pdf]
```

and this stylesheet:

```
pageSetup:
    firstTemplate: myLayout
    margin-bottom: 8mm
    margin-left: 12mm
    margin-right: 12mm
    margin-top: 8mm

pageTemplates:
    myLayout:
        frames:  [[0cm, 0cm, 100%, 100%]]
        defaultHeader =
            Header: Section: ###Section###, Title: ###Title###, Page ###Page###
        defaultFooter =
            Footer: Section: ###Section###, Title: ###Title###, Page ###Page###
        showFooter: true
        showHeader: true
```


I get the correct information in the header and footer as seen in the attached PDF.

(https://github.com/rst2pdf/rst2pdf/files/2103864/test_header_footer_in_stylesheet.pdf)

akrabat:
@jmones Are you able to provide a small Sphinx project that shows the problem?

lornajane:
Closing due to inactivity and we think it's improved but please re-open and comment if there's more to do here!

PR #653, opened 2018-04-24
#652 Set title and author correctly so that ###Title### in header/foo…

jmones:
…ter work again

This pull request is a quick fix for #652 https://github.com/rst2pdf/rst2pdf/issues/652


jmones:
Apologies for my lack of response. Unfortunately I cannot look at this now.

PR #657, opened 2018-04-27
Make exceptions py3 ready

ralsina:
* Make except statements py3-compatible 
* Fail to run setup.py at all if py3 is used.


oz123:
@ralsina  I have also added a classifier in  #658

PR #659, opened 2018-04-30
Honour the Sphinx today config setting

akrabat:
Sphinx has a `today` setting that should be used instead of the current
date if it exists. Use this for the date on the cover page.

This has the useful side-effect of making it possible to use a cover page in the tests too!


akrabat:
Hopefully, we'll really have a successfully passing Travis now!

PR #661, opened 2018-05-02
Display the reason why a test was skipped

akrabat:
When running `nosetests` with the `-v` switch, display the reason that a test was skipped by reading it from the `.ignore` file.


akrabat:
If anyone gets a moment, could they run their eyes over this & check that I haven't missed anything obvious?

Issue #662, opened 2018-05-02
Line numbers are incorrect


akrabat:
`test_issue_310` is failing. This needs to be investigated and solved.

ralsina:
The code blocks support in rst2pdf predates reSt's so it could maybe just be deleted.

akrabat:
This may be a case of me not understanding the test!

```
.. code-block:: text
   :linenos:
   :include: include
   :linenos_offset:
   :start-at: Line 3
```

I didn't realise that you could put the text of the line, rather than the line number in `start-at`…

ralsina:
What happens if you start at line 0? It may just be a case of an off-by-one :-)

akrabat:
There is no `Line 0` in [include](https://github.com/rst2pdf/rst2pdf/blob/master/rst2pdf/tests/input/include)…

akrabat:
The other problem is:

```
.. code-block:: text
   :linenos:
   :include: include
   :linenos_offset:
   :start-after: Lin
```

produces line numbers that are off by one too:

![screen shot 2018-05-03 at 08 45 47](https://user-images.githubusercontent.com/33135/39565043-64b945e2-4eae-11e8-921f-3866929fa90e.png)

PR #663, opened 2018-05-03
Fix start-at line numbers

akrabat:
In test_issue_310, the line numbers for the :start-at: test when using :linenos_offset: are off by one.

Addresses one of the problems with #662.


akrabat:
Note, that as the `test_issue_310.ignore` file is still in place, the Travis tests don't actually check that this half of #662 fixed…

lornajane:
This looks good to me, the line numbers are fixed and my output matches the reference doc.  I get a new md5 though, that you might like to add?  It's `939b8f9fa1a592df9221c15657088bff`.

akrabat:
Thanks @lornajane!

PR #671, opened 2018-06-25
Use current working directory for image path

akrabat:
Change the search for images algorithm to look for images in paths relative to the current working directory first before falling back to paths relative to the directory that the input file is in.

i.e. given `files/test.rst` of:

```rst
.. image:: images/foo.png
```

then:

```
$ rst2pdf files/test.rst -o output.pdf
```

will currently look for `foo.png` in `files/images/` with today's rst2pdf. However, after this PR, it will look in `./images/` first before falling back to  `files/images/`.

Builds on and replaces https://github.com/rst2pdf/rst2pdf/pull/574:

* Rebase to current master
* Ensure `.. image:` works in addition to background images
* Enable style sheets to be found relative to directory of input file
* Add tests


lornajane:
I'm not sure this makes sense. If defining a path on the command line, then the path should be relative to the current working directory when the command was run. But if the path is in a file, as it is for images or fonts, I would expect the referenced assets to be relative to the file where they are defined.

If I've understood this correctly then the output of `rst2pdf files/test.rst` could differ from the output of `cd files && rst2pdf test.rst`, which I would find surprising.

If the paths are defined in the stylesheet, I think I'd expect the referenced file to be relative to the stylesheet rather than the source file. Usually mine are all in one directory though so I haven't run into these problems!

akrabat:
Yes. This is definitely a situation I haven't ever encountered either.

I could imagine a scenario where you have this directory layout though:

```
.
├── dark
│   └── images
│       └── flowchart.png
├── light
│   └── images
│       └── flowchart.png
└── src
    └── presentation.rst
```

if `presentation.rst` referenced the image as `images/flowchart.png` then you could get the light and dark versions of the image by just calling `rst2pdf` from either the `light` or `dark` directory. 

> If the paths are defined in the stylesheet, I think I'd expect the referenced file to be relative to the stylesheet rather than the source file. Usually mine are all in one directory though so I haven't run into these problems!

This is a good point too. In the example above, I could imagine having a `src/presentation.style` with `light/colours.style` and `dark/colours.style` for specialism. Presumably, I'd want a background image defined in `src/presentation.style` to reference the current working directory. At the moment, I think it references from the directory that the input file is stored in, but I haven't tested.


Whether we want to support this use-case is one question. If we do, should the search path be "relative to input file, then relative to current working directory" or "relative to current working directory, then relative to input file" ?

lornajane:
I think we can cover a bunch of use cases such as:
* image is defined in main RST source and the file path is relative to that source file (mainstream use case)
* image is defined in stylesheet and the file path of the image is relative to the source file (something like a background image)
* image is defined in stylesheet and the file path of the image is relative to the stylesheet itself (could be a logo, fonts belong here too IMO)
* image is defined in the source file and file path is relative to the stylesheet (your colour-sensitive diagram example fits here)

What seems logical to me is to start with the most specific option: path search starts with relative to the source file (specific to this one document), then relative to the stylesheet (presumably might be applied to a few different documents).  I am not sure that it ever makes sense to be relative to the command being run but there's no reason NOT to support this in case it could be useful in some other situation that we haven't thought of.

akrabat:
Closed this one. Tracking the better solution here: https://github.com/rst2pdf/rst2pdf/issues/672

PR #674, opened 2018-06-26
Sphinx repeat table rows

akrabat:
Support the `repeat-table-rows` option in Sphinx.

Builds on #592 by rebasing and adding tests.


lornajane:
The contents of this pull request look OK to me - the change is plausible, the tests pass and the output looks like the title of the pull request suggests.  Although I think I'm doing something wrong because these tests seem to pass even on the master branch version of `rst2pdf` ....

I'm unclear how this feature looks from a user point of view.  Is it a command line switch (it looks like it is) - so do we need to update some documentation and the output of `rst2pdf --help` ?  What is the sphinx test doing?  Is this something I'd come across as a user or some internal thing?  Is there a way I could specify this in my source file rather than switching it at runtime?

akrabat:
> The contents of this pull request look OK to me - the change is plausible, the tests pass and the output looks like the title of the pull request suggests. Although I think I'm doing something wrong because these tests seem to pass even on the master branch version of `rst2pdf` ....

The `tests/input/test_repeat-table-rows.txt` should pass on master as I created a test as there wasn't one for `--repeat-table-rows`.

The Sphinx test should fail on master though.

> I'm unclear how this feature looks from a user point of view. Is it a command line switch (it looks like it is) - so do we need to update some documentation and the output of `rst2pdf --help` ? What is the sphinx test doing? Is this something I'd come across as a user or some internal thing? Is there a way I could specify this in my source file rather than switching it at runtime?

`--repeat-table-rows` is documented in https://github.com/akrabat/rst2pdf/blob/sphinx-repeat-table-rows/doc/rst2pdf.rst - I think that's the man page and is automatically pulled into the pdf manual?

The Sphinx test adds a new option to `pdfbuilder` (The Sphinx plugin) called `pdf_repeat_table_rows` which can then be set in `conf.py`. This wasn't documented, but is now!

Issue #544, opened 2015-07-17
AttributeError: ParaFrag instance has no attribute 'width'


belate:
Please could someone help me with this error?

[ERROR] pdfbuilder.py:130 ParaFrag instance has no attribute 'width'
Traceback (most recent call last):
  File "/Users/belate/.virtualenvs/sphinx/lib/python2.7/site-packages/rst2pdf/pdfbuilder.py", line 127, in write
    docwriter.write(doctree, destination)
  File "/Users/belate/.virtualenvs/sphinx/lib/python2.7/site-packages/docutils/writers/**init**.py", line 80, in write
    self.translate()
  File "/Users/belate/.virtualenvs/sphinx/lib/python2.7/site-packages/rst2pdf/pdfbuilder.py", line 621, in translate
    compressed=self.compressed)
  File "/Users/belate/.virtualenvs/sphinx/lib/python2.7/site-packages/rst2pdf/createpdf.py", line 653, in createPdf
    pdfdoc.multiBuild(elements)
  File "/Users/belate/.virtualenvs/sphinx/lib/python2.7/site-packages/reportlab/platypus/doctemplate.py", line 1014, in multiBuild
    self.build(tempStory, **buildKwds)
  File "/Users/belate/.virtualenvs/sphinx/lib/python2.7/site-packages/reportlab/platypus/doctemplate.py", line 927, in build
    self.handle_flowable(flowables)
  File "/Users/belate/.virtualenvs/sphinx/lib/python2.7/site-packages/rst2pdf/createpdf.py", line 762, in handle_flowable
    if frame.add(f, canv, trySplit=self.allowSplitting):
  File "/Users/belate/.virtualenvs/sphinx/lib/python2.7/site-packages/rst2pdf/flowables.py", line 559, in add
    return Frame.add(self, flowable, canv, trySplit)
  File "/Users/belate/.virtualenvs/sphinx/lib/python2.7/site-packages/reportlab/platypus/frames.py", line 167, in _add
    w, h = flowable.wrap(aW, h)
  File "/Users/belate/.virtualenvs/sphinx/lib/python2.7/site-packages/reportlab/platypus/paragraph.py", line 1091, in wrap
    blPara = self.breakLines(self._wrapWidths)
  File "/Users/belate/.virtualenvs/sphinx/lib/python2.7/site-packages/reportlab/platypus/paragraph.py", line 1358, in breakLines
[...]

PR #675, opened 2018-07-08
Add a test for the case documented in #544

lornajane:
A reportlab bug causes a layout problem when two consecutive lines have the same line length.  This is fixed in reportlab v3.3.8.

The issue is fixed upstream, we don't need to do anything.  I thought it might be useful to have a test for this scenario rather than only having the information in an old bug report though!


rst2pdf/tests/md5/test_reportlab_issue_81.json
```diff
@@ -0,0 +1,14 @@
+bad_md5 = [
+        'sentinel'
+]
+
+good_md5 = [
+        '5f323a67b1b8db8e89de364c82d34fca'
```
akrabat: Missing comma here which I think is making Travis fail.

PR #686, opened 2018-09-23
Some failing tests produce visually identical output on my platform

lornajane:
I added some good hashes for tests that "fail" but where the PDFs are exactly the same as the reference ones.


ralsina:
Just as a sidebar: there are some cases where the test is not purely visual (for example, PDF TOC issues) perhaps those should be marked as requiring a manual check beyond the visual one.

akrabat:
> Just as a sidebar: there are some cases where the test is not purely visual (for example, PDF TOC issues) perhaps those should be marked as requiring a manual check beyond the visual one.

Good point. Do you mean cases such as links?

(It's _so_ annoying that different versions of ReportLab generate different PDFs that are the "same"!)

PR #687, opened 2018-09-25
Tool to generate new "good" md5s for tests with an updated reportlab (or other changed dependencies)

lornajane:
The problem is: when a new version of reportlab is released, basically every test will fail because the files change slightly and so the checksum that we use to check the correct file was produced doesn't match.  For example: using everything matching from `requirements.txt` but the current reportlab version of 3.5.6 I get 296 tests, 23 skipped but 253 failed.

This script aims to produce PDFs and compare the outputs "visually" to see if they really look the same.  The "visually" bit is actually imagemagick's compare utility which can give a measure of difference where 0 is no differences. If the test fails but then the PDFs "look" the same, the script `migrate_checksums.py` script will run the test again with `-u good` to store the resulting md5.  I now have 296 tests, 23 skipped and 103 failures - so it's not a silver bullet.  The md5 updates are included in this PR but as a separate commit, I wasn't sure if it was helpful.

The script also attempts to create a visualisation of the differences, these are the `.jpg` files that appear in the output directory. I _think_ we could probably go to comparisons of < 0.2 or something but I haven't looked at any nonzero results yet.

This is probably work in progress to see if we can identify more similar-enough-to-be-acceptable outputs but I wanted to share where I got up to and see what people think!


akrabat:
This is fantastic and I love the approach.

However, it fails on PDFs with multiple pages, so I think we need to break into separate pages first and then compare.

i.e. comparing the multi-page PDF gives a difference of 152:

```
$ compare reference/test_issue_310_2.pdf output/test_issue_310_2.pdf  -compose src -metric PHASH output/test_issue_310_2-differences.jpg
152.265
``

Comparing separate pages shows that both pages are identical:

```
$ convert reference/test_issue_310_2.pdf reference/test_issue_310_2.png
$ convert output/test_issue_310_2.pdf output/test_issue_310_2.png
$ compare reference/test_issue_310_2-0.png output/test_issue_310_2-0.png -compose src -metric PHASH output/test_issue_310_2-differences.jpg
0
$ compare reference/test_issue_310_2-1.png output/test_issue_310_2-1.png -compose src -metric PHASH output/test_issue_310_2-differences.jpg
0
```

akrabat:
Other things:

1. Remove the old usage - we just have one usage: the new way!
2. Document the dependencies on Ghostscript and ImageMagick in the comments at the top
3. Support passing in the name of a test. i.e. `./migrate_checksums.py input/test_issue_310_2.text`

oz123:
This is indeed a fantastic approach!

ralsina:
I am going to be a bit annoying here. I LOVE THIS.

I love it so much I hate it being here since september. So I am going to merge it.

If something breaks slightly or could be done differently, let's iterate with this already merged in :-)

lornajane:
Well I'm taking this as a huge compliment :) I do think though that keeping something like this in a PR works well because it keeps all our discussions in one place. I'll have to find this closed pull request again next time I am ready to take another bite at this task - it's certainly not really ready to be used in anger.

ralsina:
@lornajane just start a new one for each papercut you want to fix. And this way if I or someone else wants to take a stab at a specific issue we just do a separate small branch.

And yes, I am a bit dogmatic about merging early and often to master :-)

lornajane:
I do agree about many small pull requests being better for changes which can stand alone. For this though, it's rather half-baked and I think could be damaging in its current state. Could we perhaps agree a way to flag a PR that is open for discussion and collaboration rather than being ready to merge? I don't want to have to work alone and in secret! Maybe a label? Or `[WIP]` in the title?

ralsina:
I usually use [WIP] when it's not ready too

akrabat:
> Could we perhaps agree a way to flag a PR that is open for discussion and collaboration rather than being ready to merge? I don't want to have to work alone and in secret! Maybe a label? Or [WIP] in the title?

Yes. `[WIP]` in the title is the way forward as non-maintainers can set that when they can't set labels.

I will create a `MAINTAINERS.md` file with some notes on my views about dealing with PRs so that we are all on the same page.

PR #689, opened 2018-09-30
Ignore .python-version

akrabat:
If you're using pyenv, you can set the python version/virtualenv that
you want to use while developing in .python-version, so this is a good
file to ignore.


oz123:
Trivial change, thanks for the update.

PR #691, opened 2018-10-05
Py23 compat additions

oz123:
Some work to make the tests run with Python3. Still not perfect, but at least we can invoke the
tests with Python3 now.


rst2pdf/rson.py
```diff
@@ -151,8 +151,12 @@ def newstring(source, client):
                 source = source.encode('utf-8')
 
             # Convert MS-DOS or Mac line endings to the one true way
-            source = source.replace('\r\n', '\n').replace('\r', '\n')
-            sourcelist = splitter(source)
+            if six.PY2:
+                source = source.replace('\r\n', '\n').replace('\r', '\n')
```
ralsina: You can use b"bytes" in python2. At least in python 2.7 which honestly is the only one worth supporting nowadays.

rst2pdf/rson.py
```diff
@@ -161,15 +165,19 @@ def newstring(source, client):
...
             # Set up to iterate over the source and add to the destination list
             sourceiter = iter(sourcelist)
-            next = sourceiter.next
-            offset -= len(next())
+            if six.PY2:
+                next_ = sourceiter.next
+                offset -= len(next_())
+            else:
+                next_ = sourceiter.__next__
+                offset -= len(next_())
```
ralsina: maybe move this line outside the if?

rst2pdf/rson.py
oz123: I don't intend to make any fancy changes to support Python2.X goal is to reach a state where the tests work on Python3. If we do this, we might be able to attract more people to actually contribute to the project, and people can use it in modern projects.

rst2pdf/rson.py
oz123: you mean offset?

ralsina:
Thanks for all the great work @oz123 !

PR #695, opened 2018-10-09
This is what a 8-year-old bug looks like

### LINKED ISSUE #612: Usig a table as a header
I'm trying to use a table like this as a header for my document:

```
+----------------------------+----------------------+----------------------+
| .. image:: images/logo.png | .. class:: headtitle | |codigo|             |                            |                           |                      |                      |
|                            | **###Title###**      +----------------------+
|                            |                      | |version|            |
|                            |                      +----------------------+
|                            |                      | |vigencia|           |
+----------------------------+----------------------+----------------------+
```
I add a image of the code for the table:

screenshot_20170322_161525

The problem is that when I convert a pdf the content of the last row of the right column is not displayed when I use it as a header for the domument, but when I use it as part of the body it works fine

Any idea what might be wrong?

ralsina:
Fix #612


akrabat:
Such a small change! I'll test is this evening.

Is there already a test that covers this?

ralsina:
No, but should be easy to add, the sample in the ticket should work as a testcase.

OTOH replaceTokens() *could* have unit tests, but the effort it took to find that sucker took about all the time I have for this.

akrabat:
I'll add a standard test for now then.

Issue #297, opened 2015-03-12
default pygments style leaks


GoogleCodeExporter:
```
If you use the bw stylesheet, which is supposed to highlight syntax without 
using colour, you still get colour.

That's because default.style is loaded first and then bw.json doesn't 
**completely** override it, so some styles come out on the other color scheme.

The solution is to generate more complete  pygment stylesheets.


```

Original issue reported on code.google.com by `roberto.alsina` on 22 Mar 2010 at 2:26

PR #700, opened 2018-11-07
[WIP] Fix 467

### LINKED ISSUE #467: Sphinx: inline syntax highlighting?
```
Is there a way I can make rst2pdf keep the inline syntax highlighting -- which 
is done using custom roles:

.. role:: py(code)
   :language: python
   :class: highlight

Test inline: :py:`for i in range(5):`. 


* Version of rst2pdf?
0.92
* Version of reportlab?
No idea.
* Are you using wordaxe, and if true, what version?
No.
* Are you using rst2pdf with Sphinx, and if true, what version?
1.1.3


```

Original issue reported on code.google.com by `kheyfbo...@gmail.com` on 29 Aug 2012 at 10:26

ralsina:
* Fix #467 (inline syntax highlight)
* Updates syntax HL stylesheets to the ones provided by newer pygment


ralsina:
So, it looks like I need to check some tests.

ralsina:
There is a regression in the styles generation that breaks issue #297

ralsina:
This PR got too old and too large, will do a separate one that fixes only #467 and add a new issue for the broken/missing pygment stylesheets.

PR #702, opened 2018-11-07
Add scale_width option

### LINKED ISSUE #505: New option for --fit-background-mode
```
Hi, for a project I work on, I must set an image as background but I only need 
to scale its width.

The attached patch adds a 'scale_width' value to the fit-background-mode option.

By the way, thanks for this great software!
```

Original issue reported on code.google.com by `ngu.anto...@gmail.com` on 26 Sep 2013 at 9:01

Attachments:
- [scale_width.patch](https://storage.googleapis.com/google-code-attachments/rst2pdf/issue-505/comment-0/scale_width.patch)

ralsina:
An actual user-originated patch!

Close #505


akrabat:
Does this need an update to the manual?

ralsina:
Looks like the ``--fit-background-mode`` option is not described in the manual, only mentioned in the option's help and man page, so added it there and ... ok, I accidentally pushed that to master. Not my smartest move. Rolling that one back and pushing here in a bit.

PR #704, opened 2018-11-08
Fix #549

### LINKED ISSUE #549: Missing scheme in --inline-links handling
As reported on https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=796959

> it seems that the handling of the --inline-links option is a bit incomplete:
> 
> https://sources.debian.net/src/rst2pdf/0.93-4/rst2pdf/genpdftext.py/#L94
> 
> only lists http:// and ftp://. Obviously missing is https, but there are other.

I've checked that this is still the case in master: https://github.com/rst2pdf/rst2pdf/blob/master/rst2pdf/genpdftext.py#L94

ralsina:
Fix #549 by ignoring all non-mailto URL schemas and passing them as-is


akrabat:
Do we need a test case for this one?

ralsina:
The fix is fairly trivial, but I will add one.

ralsina:
Added directly in master

PR #705, opened 2018-11-08
Fix #547

### LINKED ISSUE #547: Blank rst file generates a PDF with zero pages
Was expecting it to generate a single, blank page.

(Alternative title: cannot generate a blank PDF..)

ralsina:
Make empty documents generate a single empty page instead of a pdf with no pages. Fix #547


akrabat:
That's a lot of failures on Travis. I wonder what changed in the build system...

ralsina:
It *could* be because of the change. There is a "secondary doctree" which is usually empty, so I need to take a real look at the failures.

ralsina:
Now that's better :-)

PR #707, opened 2018-11-09
Kill tenjin

### LINKED ISSUE #696: remove tenjin? replace?
The code base also has a template engine Tenjin.
It's seems that upstream has a version ported to Python3.2

Also, I still don't understand what is it used for. Maybe @ralsina  can explain in a sentence or two.

ralsina:
Fix #696 

* Remove Tenjin
* Use Jinja2
* Fix all shipped templates

This is invasive and not backwards-compatible, since the template syntax is slightly different.


ralsina:
Not there yet ;-)

akrabat:
I suspect that this needs rebasing.

oz123:
Awesome, another little obstacle for python 3 removed.

PR #709, opened 2018-11-14
Add MAINTAINERS.md

akrabat:
An initial set of guidelines for maintainers to ensure that we're all on the same page.

There's no intention to criticise anyone at all, I just want to ensure that we all work the same way and have the same expectations.

Comments and improvements welcome!


oz123:
@akrabat, you have my blessing for this document.

PR #715, opened 2018-11-19
Update reportlab to 3.5.10

ralsina:


ralsina:
So, newer ImageMagick refuse to compare PDFs:

```
$ compare output/test_issue_303.pdf reference/test_issue_303.pdf 
compare: attempt to perform an operation not allowed by the security policy `PDF' @ error/constitute.c/IsCoderAuthorized/408
```

Looks like this involves creating a configuration file allowing that, looking into it.

ralsina:
Sigh, it involves editing ``/etc/ImageMagick-7/policy.xml`` and removing the PDF policy. Even setting up your own config file doesn't work because you can't override that. Pain.

akrabat:
@ralsina Fixed up some hashes for some of the tests that were failing. Many more to do though.

ralsina:
Baby steps!

ralsina:
After #716 merges we get a bunch more :-)

akrabat:
I see that #717 picked up a set that I'd done manually !

ralsina:
eeek, nasty conflicts. I will take care of them later today.

ralsina:
82 failures left.

ralsina:
Down to 50 failures. A few seem to be caused by differing outputs from inkscape, which I can't quite reproduce since our CI server installs an ancient version.

ralsina:
DONE. The only remaining failures are caused by a different version of inkscape. I expect they are ok, but since I can't obtain the artifacts, I have no way to be sure.

akrabat:
I'll sort out how to grab the artefacts and have a look.

akrabat:
_test_foreground_ positions the Confidential text differently.
(Travis version on left, reference on right)

screenshot 2018-12-05 at 08 52 24

akrabat:
On Travis, test_issue_311 has the wrong relative link as clicking on it gives me this message:

> The link ”file:///home/travis/build/akrabat/rst2pdf/rst2pdf/tests/input/images/biohazard.pdf” cannot be opened because the destination file cannot be found.

[test_issue_311.pdf](https://github.com/rst2pdf/rst2pdf/files/2647989/test_issue_311.pdf)
[test_issue_311.log](https://github.com/rst2pdf/rst2pdf/files/2647991/test_issue_311.log)

akrabat:
Just test_foreground and test_issue311 to go! Good work @ralsina  :)

ralsina:
@akrabat I think the test_foreground difference is not really important. It's there, it's in the foreground :-)

As for 311, that is probaly because to make the link work you need to be in input/ and have the input/images/ folder there, since it's a relative link. I suspect it's not even a feature worth having except for completeness since PDFs are supposed to be self-contained, and relative links are almost guaranteed to be bad.

akrabat:
Why gitlab?

ralsina:
It's supposed to support junit tests (to see individual test results) and artifacts (to get the actual output files from CI runs)

ralsina:
@akrabat that costs, like, money :-)

ralsina:
We can ignore the gitlab check, because it's just a different system generating slightly different checksums in a few tests (I think because of inkscape). Removing the [WIP] from this PR, I say merge at will :-)

akrabat:
oh, it passed!

akrabat:
Thanks for your hard work on this @ralsina!

PR #717, opened 2018-11-28
Fix #716

### LINKED ISSUE #716: We are missing reference artifacts
There are a number of test cases where:

1. The test passes in master
2. We have no reference artifact

That means we can't use @lornajane 's script to automatically check for visual differences when upgrading reportlab.

So: one-time-thing, I will do a PR to add reference artifacts in those cases.

Additionally, we have artifacts without a known good checksum!

```
BAD ARTIFACT: unknown  reference/test_issue_349.pdf
BAD ARTIFACT: unknown  reference/test_issue_175.pdf
BAD ARTIFACT: unknown  reference/sphinx-issue251.pdf
```

I will check those manually and add the checksum if they are ok, or remove the artifact instead.

ralsina:
* Add reference artifacts that are known good from master
* Remove / validate reference artifacts that are NOT known good in master


akrabat:
Add me as a reviewer when you're ready for me to look?

ralsina:
@akrabat there you go :-)

rst2pdf/tests/check_artifacts.py
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env python
+
```
akrabat: Can we add a comment here about what this script does and how to use it?

rst2pdf/tests/check_artifacts.py
```diff
@@ -0,0 +1,29 @@
...
+    md5_path = artifact.replace('.pdf', '.json').replace('reference/', 'md5/')
+    res, _ = checkmd5(artifact, md5_path, [], False)
+    if res != 'good':
+        print 'BAD ARTIFACT: %s ' % res, artifact
+
+# Check if any output file is good and has no matching artifact
+
+for new_artifact in glob.glob('output/*.pdf'):
+    artifact = new_artifact.replace('output/', 'reference/')
+    md5_path = artifact.replace('.pdf', '.json').replace('reference/', 'md5/')
```
akrabat: We only need to calculate `md5_path` if the artifact doesn't already exist in the reference folder. So maybe move this line into the `else` statement?

PR #718, opened 2018-12-13
Update requirements

akrabat:
This picks up an update to urllib3 which had a security issue prior to 1.23.

(also updates ReportLab to 3.5.12, but this doesn't seem to be a problem!)

Can only upgrade Sphinx to 1.7.9 because 1.8+ causes failures (See #721 & #722)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,8 @@ addons:
   apt:
     packages:
     - inkscape
+    - texlive-latex-base
+    - dvipng
 
 language: python
 python:
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,8 @@
 {next}
 ------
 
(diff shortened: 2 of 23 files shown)


oz123:
@akrabat any idea why is the test failing?

akrabat:
Nope. It passed on my computer...

akrabat:
ah - I can reproduce now - I can't operate pip install!

akrabat:
sphinx-issue168 failure is genuine - `linenothreshold` has stopped working…

akrabat:
I think this is ready to review. In particular, note https://github.com/rst2pdf/rst2pdf/pull/718/commits/6d76e69c3bc494669b1fdb1b3f60598cb31b71f3

oz123:
Awesome works. I wonder one can make the rest less fragile. But I guess that not easy with any other large software project.

PR #726, opened 2019-01-03
Fix #725

### LINKED ISSUE #725: Remove uniconvsaver.py
Looks like nothing is using it anymore since #487

ralsina:
Remove unused file (Fix #725)

Ouch, looks like I branched out of the kill-smarty branch, so only consider this after that one is merged, I guess.


akrabat:
> Ouch, looks like I branched out of the kill-smarty branch, so only consider this after that one is merged, I guess.

Rebased against master.

PR #729, opened 2019-01-09
Some code cleanups

ralsina:
Fixes a few issues reported by flake8, as in #724. Not all, not a lot, but some :-)


ralsina:
@oz123 why not!

PR #737, opened 2019-02-11
More cleanups 3

ralsina:
Only merge after more-cleanups-2 :-)


oz123:
I an a way I think it would be a bit nicer to just squash the commits in this PR and then title this commit pepi8y.

ralsina:
Just "squash and merge"

PR #747, opened 2019-02-13
Migrate to svglib

### LINKED ISSUE #745: Deprecate svg2rlg?
It's py2-only, not widely used, and inkscape gives much better results (although yes, it's a pain in OSX)

ralsina:
* svglib has python3 support
* it has actual releases
* it's actively maintained

Fix #745


ralsina:
The remaining tests that fail in Gitlab fail differently on repeated runs, so are a bit intractable right now.

The one remaining failure in Travis is probably ok but I would need to see the artifact to check.

ralsina:
Noticed GL-CI was using python 3.5... that is the last python 3 version that has random-ordered dictionaries, which could be causing our trouble with broken invariancy. Updated GL config to 3.6 and if that's good can mark that as required minimum python 3 and all will be ok.

PR #755, opened 2019-02-23
Add frame padding configuration to pageTemplate frame config (Fixes #753)

### LINKED ISSUE #753: Unexplained padding on images results in new page
Given `presentation.rst`:

```rst
.. image:: coffee.jpg
    :height: 9cm

.. raw:: pdf

   PageBreak aPage

.. image:: coffee.jpg
    :height: 9cm
```

with `slides.style`:

```
pageSetup:
    width: 16cm
    height: 9cm
    margin-top: 0cm
    margin-bottom: 0cm
    margin-left: 0cm
    margin-right: 0cm
    margin-gutter: 0cm
    spacing-header: 0cm
    spacing-footer: 0cm
    firstTemplate: aPage

pageTemplates:
    aPage:
        frames: []
            [0cm, 0cm, 100%, 100%]
        showHeader : false
        showFooter : false
```

and I create a PDF using:

```
rst2pdf -s slides.style presentation.rst -o presentation.pdf
```

I get a PDF with three pages and the two image pages do not have the image full screen.

preview-of-pdf

Where is this padding/margin coming from and how do we remove it?

Coffee image here: https://akrabat.com/stuff/coffee.jpg

ralsina:
Allow 4 extra element in a frame definition to specify frame padding (defaults to 6pt for backwards compatibility)

Fixes #753


akrabat:
Solves the problem!

Question: Should we make the default for the frame padding `0` ?

akrabat:
Needs a test file too

ralsina:
Maybe but it will break every test

akrabat:
That's the thing, isn't it?!

ralsina:
So, a more extensive test shows this is still buggy, so marking this as WIP

ralsina:
The code was correct, and the test was wrong :-D

ralsina:
@akrabat I am willing to do all the test checking for setting it to 0 but it's badly backwards-incompatible ... your call :-)

rst2pdf/tests/input/test_753.txt
```diff
@@ -0,0 +1,27 @@
+.. image:: images/biohazard.png
```
akrabat: Can we add a page before this page in the PDF that explains what the next pages should look like as that'll make it easier for future people to confirm that the test is passing?

akrabat:
The backwards compatibility thing is a real issue. Technically as we're pre-1.0, BC is okay, but equally, we've been the same way for many years!

Anyone else have an opinion?

rst2pdf/tests/input/test_753.txt
ralsina: Done!

PR #758, opened 2019-03-04
Docs: We use svglib

flying-sheep:
As of #747

reportlab also already supports gradient, but svglib doesn’t yet:

- https://github.com/deeplook/svglib/issues/50
- https://github.com/deeplook/svglib/issues/142


oz123:
@flying-sheep can you please squash both commits?

PR #763, opened 2019-03-12
Remove lazy imports

ralsina:
Remove LazyImports class.

Yet another step in my quest to make rst2pdf's codebase less weird :-)

### DIFF
--- a/rst2pdf/opt_imports.py
+++ b/rst2pdf/opt_imports.py
@@ -108,3 +108,5 @@ class LazyImports(object):
     #     return None
 
 LazyImports = LazyImports()
+
+from PIL import Image as PILImage
--- a/rst2pdf/opt_imports.py
+++ b/rst2pdf/opt_imports.py
@@ -87,12 +87,12 @@ class LazyImports(object):
         setattr(self, name, value)
         return value
 
-    def _load_pdfinfo(self):
-        try:
-            from pyPdf import pdf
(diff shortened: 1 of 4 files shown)


rst2pdf/opt_imports.py
```diff
@@ -69,42 +69,5 @@
...
-
-    def _load_svg2rlg(self):
-        if six.PY2:
-            import svg2rlg
-            return svg2rlg
-        return None
-
-LazyImports = LazyImports()
+import pdfrw as pdfinfo
+from PIL import Image as PILImage
```
akrabat: What is the benefit of importing these two within opt_imports compared to importing directly in `image.py`? 

Or, to put it another way, should there be a `try` around these imports?

ralsina: Just wanted to make the change minimal.

They don't need a try/except because we are really requiring them now :-)

akrabat: Got it. My gut feeling is that they shouldn't be in opt_imports as they aren't optional. Your call though as you know this better than I do.

ralsina:
I don't want to have any optional imports at all, so in a followup branch I
can just get rid of that file

akrabat:
> I don't want to have any optional imports at all, so in a followup branch I can just get rid of that file

Works for me! 👍

PR #765, opened 2019-03-13
Improve findfonts.py

### LINKED ISSUE #748: After polyglot: fix findfonts.py
That file is pretty messy. It has a lot of confusing bytes/unicode casting, a very old-fashioned directory tree traversing code that can be replaced with simple globbing, and some other similarly bad stuff.

Since tests seem to cover it pretty well, fixing it should not be horribly hard.

ralsina:
This branch fixes one small bug in windows and makes the code a bit less complex. Should cause no change in behaviour.

Fixes #748


rst2pdf/findfonts.py
```diff
@@ -179,26 +172,25 @@ def findFont(fname):
...
 
-def findTTFont(fname):
 
+def findTTFont(fname):
     def get_family(query):
-        data = subprocess.check_output(["fc-match", query])
+        data = make_string(subprocess.check_output(["fc-match", query]))
         if six.PY2:
-            data = data.decode('UTF-8')
+            data = data.decode("UTF-8")
```
akrabat: As `make_string` does a utf-8 decode, we don't need this `if six.PY2:` statement I think.

rst2pdf/findfonts.py
```diff
@@ -179,26 +172,25 @@ def findFont(fname):
...
             if family:
                 return family
         return None
 
     def get_fname(query):
-        data = subprocess.check_output(["fc-match", "-v", query])
+        data = make_string(subprocess.check_output(["fc-match", "-v", query]))
         if six.PY2:
-            data = data.decode('UTF-8')
+            data = data.decode("UTF-8")
```
akrabat: As `make_string` does a utf-8 decode, we don't need this `if six.PY2:` statement I think.

akrabat:
By introducing psfonts into the tests, does the GPL license now apply to the entire project? I'm concerned that it might very well do that.

ralsina:
On Tue, Mar 19, 2019 at 5:39 PM Rob Allen <notifications@github.com> wrote:

> By introducing psfonts into the tests, does the GPL license now apply to
> the entire project? I'm concerned that it might very well do that.
>
> No, this is the "mere aggregation" part of the GPL.
Of course if there is a BSD postscript font out there we can use that
instead.

rst2pdf/findfonts.py
ralsina: agreed

PR #771, opened 2019-05-01
Fixed reportlab import

zacthespack:
**Describe the bug**
When running html generation I ended up with the error:

```
Exception occurred:
File "/usr/local/lib/python2.7/dist-packages/rst2pdf/flowables.py", line 895, in <module>
   if reportlab.Version == '2.1':
 NameError: name 'reportlab' is not defined
```

**To Reproduce**
```
sphinx-build -b html -d stage/doctrees   source/welcome stage/html
```

### DIFF
--- a/rst2pdf/flowables.py
+++ b/rst2pdf/flowables.py
@@ -5,6 +5,7 @@ __docformat__ = 'reStructuredText'
 
 from copy import copy
 import re
+import reportlab
 
 from reportlab.platypus import *
 from reportlab.platypus.doctemplate import *


akrabat:
I wonder why none of our current tests find this.

zacthespack:
I would agree, perhaps a test needs changing to? but as far as my usage goes this was a change that was very much needed

akrabat:
@zacthespack Are you able to create a super simple project (presumably as a zip file) that causes the problem?

invizbox:
If it helps with reproducing, I get the issue with reportlab 3.5.20 and it goes away when I set the version back to 3.5.19. I believe the Version setting is probably gone (typically replaced by __version__ as it is standard). This sounds like the recent issue with Pillow 6.0.0 and a future proof fix may be to check all places where <package>.Version is called within the codebase as I expect most packages to be migrating to __version__.

akrabat:
> If it helps with reproducing, I get the issue with reportlab 3.5.20 and it goes away when I set the version back to 3.5.19. I believe the Version setting is probably gone (typically replaced by `__version__` as it is standard). This sounds like the recent issue with Pillow 6.0.0 and a future proof fix may be to check all places where` <package>.Version` is called within the codebase as I expect most packages to be migrating to `__version__`.

I can't work out how to reproduce on 3.5.20 locally. Can you let me know your environment, python version etc?

akrabat:
I don't know exactly what I did, but I could reproduce this on a presentation I'm currently working on. 
Thanks for the fix @zacthespack.

PR #774, opened 2019-05-15
Import reportlab into flowables.py

akrabat:
This fixes #771 for 0.94.x as it's breaking on latest ReportLab and will allow us to release 0.94.1 with just this fix in it.

I've also created a `0.94.x` branch from the `0.94` tag in order to give us somewhere to release 0.94.1 from.


ralsina:
LGTM

PR #782, opened 2019-06-12
Replace deprecated BuildEnvironment.create_index()

svenschwermer:
This function was deprecated in sphinx 1.6 and removed in 2.0. The
function call is replaced with the recommended replacement according to
https://www.sphinx-doc.org/en/stable/extdev/index.html

Signed-off-by: Sven Schwermer <sven.schwermer@disruptive-technologies.com>


akrabat:
Thank you @svenschwermer! We really appreciate it.

PR #788, opened 2019-07-17
Sphinx3 compatibility fixes

manfield:
NoUri was moved under sphinx.errors
ustrftime is deprecated, used time.strftime instead

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,7 @@
 {next}
 ------
 
+* Fixed: Support Sphinx 2.1+ on Python 3 (Issue #788)
 * Changed: Support docutils math directive instead of rst2pdf's (Issue #722)
 * Fixed: HTTPS images using image directive now works with Python 3 (Issue #780)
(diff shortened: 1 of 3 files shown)


akrabat:
The CI tests are failing with:

```
 Could not import extension rst2pdf.pdfbuilder (exception: cannot import name NoUri)
```

I guess we need to dynamically import based on which version of Sphinx is in use?

akrabat:
Something like this should work:

```python
if sphinx.__version__ >= '2.1':
    from sphinx.errors import NoUri
else:
    from sphinx.util import ustrftime
    from sphinx.environment import NoUri
```

However, when I do `pip install --upgrade Sphinx==2.2`, I get two test failures, so they need looking at.

akrabat:
I think we'll take this change as the test failures look to be a difference due to Sphinx 2 vs 1.8 and as we don't yet have the ability to test different dependency versions, it would be better to have some support for Sphinx >=2.1 than none at all.

PR #792, opened 2019-08-17
[WIP] The great style fix of 2019 (+ Python 3 stuff)

stephenfin:
I really don't like applying style checkers to existing code bases, because they
tend to make a complete mess of our ability to use 'git-blame' and related tools
effectively. With that said, the code base is in pretty bad shape owing to its
sheer age and we'd otherwise be fixing this stuff for years. Bite the bullet and
run all the code through three tools - black, autoflake8 and isort - meaning we
don't have to worry about this going forward.


akrabat:
It's when I see a PR like this, that I'm really glad that we have tests!

stephenfin:
Yup 😅 I've tried to break it up as much as possible but you should definitely consider this WIP until I fix everything up. Tests are currently failing on master for me locally  (Fedora 30) so I need to figure that out first.

Wikiwide:
Will be glad when rst2pdf will be working in Python3. Because I am trying to use Sphinx 2.2 instead of 1.8.5, in attempt to get Table-of-Contents links working in PDF file generated by rst2pdf - and Sphinx2.2 is only available with Python3, not Python2. Not sure if it will help my issue with ToC, but...

Wikiwide:
Current master branch (0.95) already works fine with Python 3 (I have tested it with my Python 3.5). No need for further Python 3 modifications?

Wikiwide:
And yes, rst2pdf 0.95 (unlike 0.94.1 release) fixes problem with Table of Contents links, thank you!

akrabat:
> Current master branch (0.95) already works fine with Python 3 (I have tested it with my Python 3.5). No need for further Python 3 modifications?

See https://github.com/rst2pdf/rst2pdf/issues/796

oz123:
This is a very courageous thing to do!!!
PRs like this tend to get rejected, because they are two big to review.

setup.py
```diff
@@ -41,15 +37,17 @@ def read(*rnames):
...
+            "templates/*tmpl",
+        ]
+    },
     include_package_data=True,
-    dependency_links=[
-    ],
+    dependency_links=[],
     install_requires=install_requires,
     tests_require=tests_require,
     extras_require=dict(
```
johnthagen: If this PR is the final step for Python 3 compatibility (or at least close enough to call it officially supported), I would also suggest that you add the [trove classifiers for Python 3](https://pypi.org/pypi?%3Aaction=list_classifiers):

```
Programming Language :: Python :: 3
Programming Language :: Python :: 3.5
Programming Language :: Python :: 3.6
Programming Language :: Python :: 3.7
```

akrabat:
I think we need to do one tool at a time with no other changes in the PR.

PR #794, opened 2019-08-18
Minor doc updates

stephenfin:
- Convert markdown docs to rST (this is rst2pdf, after all)
- Some minor formatting fixes


README.rst
```diff
@@ -1,5 +1,10 @@
-rst2pdf: Use a text editor. Make a PDF.
-========================================
+=======
+rst2pdf
+=======
+
+------------------------------
+Use a text editor. Make a PDF.
+------------------------------
```
akrabat: Moving all the heading markers down a level will affect https://pypi.org/project/rst2pdf/. Are we okay with that?

stephenfin: I think it shouldn't matter. You can choose whatever symbols you like, in any order, so long as each level has a unique character.

PR #800, opened 2019-09-17
Make it still work with Python 2

Wikiwide:
Unless difference between Import Error and Module Not Found Error is that important for SVGImage import.


akrabat:
Thank you!

PR #807, opened 2019-09-26
Fix for fonts in Windows

### LINKED ISSUE #806: Sphinx builder execution failing in Windows
After installing from `master` and running Sphinx 2.2.0 `make pdf`, an exception is thrown from https://github.com/rst2pdf/rst2pdf/blob/25713408cbe8da5be0f12f1176b11fddd1d95774/rst2pdf/findfonts.py#L225

**Description of problem**
- Install rst2pdf and configure a Sphinx 2.2.0 project
- Run `make pdf`
- Build fails at `findfonts.py` (see above)

🖥 **Versions**
python 3.6.6
Sphinx 2.2.0

Windows 10

**Additional information**
This prevents all Windows users from using rst2pdf

haschdl:
This was enough to make pdfbuilder work with Sphinx 2.2.0 in Windows 10.

Fixes #806 

it_aint_much

### DIFF
--- a/rst2pdf/findfonts.py
+++ b/rst2pdf/findfonts.py
@@ -222,7 +222,7 @@ def findTTFont(fname):
         # ctypes with EnumFontFamiliesEx
 
         def get_nt_fname(ftname):
-            import _winreg as _w
+            import winreg as _w
 
             fontkey = _w.OpenKey(
                 _w.HKEY_LOCAL_MACHINE,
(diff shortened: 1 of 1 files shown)


rst2pdf/findfonts.py
```diff
@@ -222,7 +222,7 @@ def get_variants(family):
         # ctypes with EnumFontFamiliesEx
 
         def get_nt_fname(ftname):
-            import _winreg as _w
+            import winreg as _w
```
akrabat: I think this will break Python 2 on Windows.

How about something like:

```
if six.PY3:
    import winreg as _w
else:
    import _winreg as _w
```
?

akrabat:
Thank you @haschdl!

PR #811, opened 2019-10-08
Fix Preprocess extension for UTF-8 in Python 2

### LINKED ISSUE #785: Preprocess extension no longer works with UTF-8 in Python 2
**Description of problem**
Given this rst file:

```rst

Test
====

.. code-block:: text

    .
    ├── src/
```

Then this using the preprocess extension will fail on Python 2:

```
$ rst2pdf test.rst -e preprocess -v
[INFO] createpdf.py:1578 Importing extension module 'vectorpdf'
[INFO] createpdf.py:1578 Importing extension module 'preprocess'
  File "/Users/rob/.pyenv/versions/rst2pdf-2/bin/rst2pdf", line 11, in <module>
    load_entry_point('rst2pdf', 'console_scripts', 'rst2pdf')()
  File "/Users/rob/Projects/python/rst2pdf/rst2pdf/createpdf.py", line 1461, in main
    add_extensions(options)
  File "/Users/rob/Projects/python/rst2pdf/rst2pdf/createpdf.py", line 1592, in add_extensions
    module.install(createpdf, options)
  File "/Users/rob/Projects/python/rst2pdf/rst2pdf/extensions/preprocess_r2p.py", line 367, in install
    data = Preprocess(options.infile)
  File "/Users/rob/Projects/python/rst2pdf/rst2pdf/extensions/preprocess_r2p.py", line 153, in __init__
    self.sourcef = DummyFile(source)
UnicodeEncodeError: 'ascii' codec can't encode characters in position 44-46: ordinal not in range(128)
```

It works in Python 3, because `DummyFile(str)` supports unicode, but doesn't in Python 2.

🖥 **Versions**
0.95.dev0

```
$ python --version
Python 2.7.15
```

judy2k:
Fixed #785.


akrabat:
@judy2k This is great. Thanks! I've updated to 

1. Render the test PDF with a font that makes the glyphs display nicely.
2. Add missing functionality to preprocess.py so that the`test_preprocess.txt` test passes.

Could you review my changes please?

rst2pdf/extensions/preprocess_r2p.py
```diff
@@ -108,12 +108,20 @@
...
         we need here.
     '''
+    def __init__(self, value):
+        self._str = value
+
     def read(self):
-        return self
+        return self._str
+
+    def strip(self):
```
judy2k: File objects don't usually have a `strip` method - that's a method on `str` etc.

I can see you've added  a `read` below - this needs to be moved further up. I'll do this now and add the commit.

judy2k:
@akrabat I think this is ready to go now, providing it passes review :)

akrabat:
Thanks @judy2k !

PR #813, opened 2019-10-11
Ignore Sphinx _build files created when testing

akrabat:
The Sphinx tests generate an _build directory that I don't believe we need, so add `_build` to `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ rst2pdf/tests/output/*.log
 .python-version
 .venv/
 dist/
+_build/
+


lornajane:
When does this directory get made? I don't have one so I'm not sure what problem I'd see this change solving

akrabat:
> When does this directory get made? I don't have one so I'm not sure what problem I'd see this change solving

You're right. We used to create `_build` file, but [stopped](https://github.com/akrabat/rst2pdf/commit/7dae1c0e7efb43dffb43c86a21e2b4a7e9c341f8#diff-99754a3ae9362e7faa194e8a0ac32859R267).

PR #814, opened 2019-10-19
Detail supported Python 3 versions

judy2k:
Update the setup file to specify that rst2pdf supports Python 3, specifically versions >= 3.6.

Might be good to ask Travis to test against those versions too, but I'll leave that decision to a maintainer.


akrabat:
We should probably test 3.6 if we say that we support it.

PR #816, opened 2019-10-20
First pass at adding snap support

### LINKED ISSUE #743: Snapcraft release
I went to a lightning talk at FOSDEM on [Snapcraft](https://snapcraft.io) and I think we need to be here and probably in other app-stores if there are any others? 

As I understand it, Snapcraft allows users to update to the latest version of rst2pdf when the version provided by apt is out of date. It seems that most users are on LTS versions of Ubuntu and so they won't get the latest rst2pdf from apt. i.e. Snapcraft decouples rst2pdf's release cycle from Ubuntu's for users.

I also wonder if it allows us to also package up our dependencies like ReportLab independently of the system's. This would ensure that a user who is using our rst2pdf _snap_ won't get any weird incompatibility issues if they updated the system ReportLab to one we don't work with.

The obvious tasks for this ticket:

1. Investigate how to create a snap and then write the code/scripts to create it.
2. Investigate how to create an account on Snapcraft for the project and get this set up.
3. Implement a way to create snaps when Travis builds _master_ so that we can have an automated "latest dev" version in the snapcraft store.
4. Document how to create a tagged version release snap for [RELEASE_PROCESS.rst](https://github.com/rst2pdf/rst2pdf/blob/master/doc/RELEASE_PROCESS.rst)

popey:
Fixes #743 

Once (if) merged, I recommend someone from the project registers an account at snapcraft.io (link top right) and then visit build.snapcraft.io, hook it up (big green button) to this git repo. That should build you a snap and publish to the store for multiple architectures.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,28 @@
+name: rst2pdf
+base: core18 
+adopt-info: rst2pdf
+summary: Use a text editor. Make a PDF.
+description: |
+  The usual way of creating PDF from reStructuredText is by going through
+  LaTeX. This tool provides an alternative by producing PDF directly using
+  the ReportLab library.
+
+grade: stable
+confinement: strict
+
+parts:
+  rst2pdf:
+    plugin: python
+    python-version: python2
+    source: .
+    override-pull: |
+      snapcraftctl pull
+      snapcraftctl set-version $(git describe --tags)
+
+apps:
+  rst2pdf:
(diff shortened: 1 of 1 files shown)


lornajane:
I should also add instructions to the README once the snap is published and before this PR is merged. Note to self!

lornajane:
I couldn't figure out how to push back to the same PR so this commit and discussion thread will continue over in #830

PR #828, opened 2019-12-30
Prepare for 0.95 release

akrabat:
Changes for releasing 0.95


akrabat:
> Are we following a specific format for what order the items go into in the `CHANGES` file? I'd put the python3 support first and possibly in capital letters! Otherwise looks sound

CHANGES were in reverse issue number. I've now reorganised so that they are grouped by type (Added, Changed, Fixed and then Removed) and only include items that users should read.

PR #829, opened 2019-12-30
Update master version to 0.96.dev

akrabat:
With 0.95 released, we update the code to 0.96.dev so that we can distinguish a bug reported against the master branch from one against the 0.95 release.

### DIFF
--- a/doc/assets/cover.tmpl
+++ b/doc/assets/cover.tmpl
@@ -9,7 +9,7 @@
 
 .. class:: centered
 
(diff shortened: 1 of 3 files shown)


akrabat:
> Usually my rst2pdf does know it's a dev version when I'm using the source, so I don't think we need this but I'm not really sure.

You are right. The fact that it's a dev build is stored in `setup.cfg`. Somehow I managed to forget that…

Updated.

PR #830, opened 2019-12-30
Add support for publishing rst2pdf as a snap

### LINKED ISSUE #743: Snapcraft release

lornajane:
Fixes #743 and replaces #816 (because I couldn't work out how to push back to that remote when I added some updates).

@popey made us a snap configuration! I've updated it now we're on python 3 and tested it locally. I'm now in the market for other snap users to test this as well and let us know how they get on. We don't yet have access to the reserved `rst2pdf` name on snapcraft.io but once I hear from @ralsina if it's not him that has it registered I'll request it and work towards releasing this as an official install channel for this project.

To test this if you are a snap user (the process is a bit different for installing a local snap than a proper published one):
- run `snapcraft` in the root of the project to pick up the snapcraft file and build the snap
- run `sudo snap install [the snap file you just made] --dangerous` the dangerous flag indicates you know you're installing unsigned random code package!
- now try the rst2pdf command and you should get a pdf file output :) If you already have an rst2pdf install, then at least on ubuntu the new executable will be in `/snap/bin/rst2pdf` and you can use that directly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ rst2pdf/tests/output/*.log
 .python-version
 .venv/
 dist/
+snap/.snapcraft
--- a/README.rst
+++ b/README.rst
@@ -64,6 +64,14 @@ The latest released version may be installed from PyPI by using
 
     $ pip install --user rst2pdf
 
+Install from Snap
+~~~~~~~~~~~~~~~~~
+
+If you are using a system that supports [snaps](https://snapcraft.io/)
+then you can install from there with::
+
+    $ snap install rst2pdf
+
 Install from GitHub
 ~~~~~~~~~~~~~~~~~~~
 
--- a/doc/RELEASE_PROCESS.rst
+++ b/doc/RELEASE_PROCESS.rst
@@ -87,3 +87,13 @@ It should contain the following:
 .. _changelog-generator: https://github.com/weierophinney/changelog_generator
 .. _Test-PyPI: https://test.pypi.org
 .. _PyPI: https://test.pypi.org
+
+
+Releasing as a Snap
+~~~~~~~~~~~~~~~~~~~
+
+1. Update the version string in ``snap/snapcraft.yml`` as desired (probably to match the new release tag)
+
+2. Run ``snapcraft`` and note the filename of the output
+
+3. Now publish (the ``rst2pdf`` namespace is associated with @lornajane's Ubuntu account) by doing ``snapcraft push --release=stable [the snape filename from the previous step]``
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,11 +1,13 @@
 name: rst2pdf
 base: core18 
-adopt-info: rst2pdf
+version: '0.96'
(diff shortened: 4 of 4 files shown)


lornajane:
Note to self: need to update the README as well once this snap is published properly.

magicmonkey:
I tried this from Lorna's instructions, on a fairly vanilla Ubuntu 19.10, and it worked fine.  I built the snap file itself (after installing multipass), and then installed it, and successfully used the installed snap to convert an RST file into a PDF as expected.

lornajane:
OK this is good, the snap is published and @magicmonkey can install it from the main snap store now. It exists! https://snapcraft.io/rst2pdf

akrabat:
When I installed and then ran `rst2pdf --version`, it reported `0.96.dev0`. We need this to be `0.96`.

I guess that it is using the `tag_build` setting in `setup.cfg` when building. Ideally, we'd be able to set the `tag_build` value that we want in `snapcraft.yaml`, but I didn't find any hint on how to do this in the docs I found. If we can't do it that way, maybe we need a build script that removes that line from `setup.cfg` before building the snap? There's no doubt another solution too that I haven't thought of.

akrabat:
Oh, can we add some details on how to create and publish the snap into either RELEASE_PROCESS.rst or a new doc?

lornajane:
The version is an rst2pdf feature, the behaviour is the same as is on the master branch since I didn't change anything about rst2pdf itself here. If we can get this PR merged to master, we'd be able to tag a new version and release a snap of that  - and I think the tagged versions return the "normal" version information.

akrabat:
> The version is an rst2pdf feature, the behaviour is the same as is on the master branch since I didn't change anything about rst2pdf itself here. If we can get this PR merged to master, we'd be able to tag a new version and release a snap of that - and I think the tagged versions return the "normal" version information.

Ah! That's a good idea. I don't think it does by default, but when we tag, we could remove the `.dev` in `setup.cfg` for the tag and then put it back for master.

PR #836, opened 2020-01-03
Set version numbers for 0.95.post1 release

akrabat:
The 0.95.post1 release consists of just the manual update. 

As per [PEP 440](https://www.python.org/dev/peps/pep-0440/#post-releases), I've chosen to do a post-release as this change does not affect the distributed software that makes up 0.95.


akrabat:
Actually, let's do this as 0.96 to avoid any confusion.

Issue #835, opened 2020-01-02
rst2pdf/0.95: unable to run tests


sandrotosi:
```
$ PYTHONPATH=. python3 setup.py test
running test
running egg_info
writing rst2pdf.egg-info/PKG-INFO
writing dependency_links to rst2pdf.egg-info/dependency_links.txt
writing entry points to rst2pdf.egg-info/entry_points.txt
writing requirements to rst2pdf.egg-info/requires.txt
writing top-level names to rst2pdf.egg-info/top_level.txt
reading manifest file 'rst2pdf.egg-info/SOURCES.txt'
writing manifest file 'rst2pdf.egg-info/SOURCES.txt'
running build_ext
test_rst2pdf (unittest.loader._FailedTest) ... ERROR

======================================================================
ERROR: test_rst2pdf (unittest.loader._FailedTest)
----------------------------------------------------------------------
ImportError: Failed to import test module: test_rst2pdf
Traceback (most recent call last):
  File "/usr/lib/python3.7/unittest/loader.py", line 154, in loadTestsFromName
    module = __import__(module_name)
ModuleNotFoundError: No module named 'rst2pdf.tests.test_rst2pdf'


----------------------------------------------------------------------
Ran 1 test in 0.000s

FAILED (errors=1)
Test failed: <unittest.runner.TextTestResult run=1 errors=1 failures=0>
error: Test failed: <unittest.runner.TextTestResult run=1 errors=1 failures=0>
```

how do you run tests?

akrabat:
```
$ pip install nose coverage
$ pip install -r requirements.txt
$ pip install -e .[tests,sphinx,images,svgsupport,aafiguresupport,mathsupport,rawhtmlsupport]
$ nosetests -v -i regulartest -i sphinxtest
```

(from [Running Tests](https://github.com/rst2pdf/rst2pdf/blob/master/doc/DEVELOPERS.rst#running-tests))

lornajane:
We already have an issue open to "make the tests less weird" #739 which I hope will help with issues like this one

akrabat:
We should make  `python3 setup.py test` do something useful - either run nosetests or output instructions on what to do.

normanlorrain:
I'm working on towards a PR that would enable use of pytest.  Not sure it's ready yet, but FYI it's here: 
[(github)](https://github.com/normanlorrain/rst2pdf).  Tested on Python 3.8.1, on Linux (Windows WSL)

[...]

Issue #739, opened 2019-02-11
Redo tests using pytest


ralsina:
nosetests is somewhat deprecated and pytest has better tooling.

It would make the tests less weird.

PR #838, opened 2020-01-07
Port tests to use pytest.

normanlorrain:
Addresses #739 and related #835 .  My first PR to this project, so I humbly submit for review.

Many changes here, but almost all consist of moving test code to a new _tests/_ folder, outside the main body of code.  This seems to be a preferred structure for pytest.  Sphinx tests are not active; I'd like to save that for a later PR.

The main change is the new _tests/conftest.py_ file.  This is a pytest plugin that adapts the existing test format to what pytest requires.  I tried to reuse code from _autotest.py_ as much as possible.  There's some dead code here, but again to keep this large PR simple I've deferred that. 

Note: I was initially using Black to auto-format code, so to simplify the PR I reverted some of my commits.  Some output PDF's were also inadvertently committed, so that was fixed too.

To run the tests, 
```
pip install pytest
pytest

```
Tested on Linux Debian (Windows WSL), Python 3.8.1.  Currently 6 tests fail, 267 pass.  Failed tests are due to md5 changes, so might or might not be acceptable.

Not tested under Windows because of underlying test code (_execmgr.py_).


oz123:
Very impressive! Travis tests fail because of matplotlib version . This needs to be updated to 3.1.2.

akrabat:
I'm a little confused to see changes to the test files themselves. Isn't this a reorganisation of where the files live and what runs the tests?

Also, I'm not sure why single quoted strings have been converted to double quotes and the matplotlib dependency changed?

normanlorrain:
> I'm a little confused to see changes to the test files themselves. Isn't this a reorganisation of where the files live and what runs the tests?
> 
> Also, I'm not sure why single quoted strings have been converted to double quotes and the matplotlib dependency changed?

I'll put the test code back to original state (quote changes, etc.) and the requirements.txt as well.  I think I can just update this PR.

normanlorrain:
I see that I made a mistake with this PR, in basing it on my fork's master branch, rather than a proper, separate branch.

Now that my fork's master has fallen behind the upstream master, I'm having trouble rebasing it.  I'm getting several conflicts, which I haven't figured out yet.  (I ended up doing `git rebase --skip`, but I'm not sure if that loses anything I've added).

I may have to restart, by forking again and recreating this on a proper branch and re-creating the PR.  

I'll leave this for now, waiting for #841 to be complete.

akrabat:
Thanks for your work @normanlorrain. I've just merged #850 where @stephenfin used your approach from here.

PR #841, opened 2020-01-20
Support Python 3 only

akrabat:
As Python 2 is no longer supported, we no longer need to support it for rst2pdf. 

This PR remove Python 2.7 from the Travis testing and adds Python 3.8. I had to update the dependencies for Python 3.8, so I updated everything by Sphinx to the latest. A number of tests needed updated hashes, which I did after visually checking the output PDFs against the reference. Note that Sphinx has not been updated the next version of Sphinx changes the text that is generated on our table of contents pages which will need investigation.

This is a prerequisite for #838.


lornajane:
I haven't forgotten about this but I can't get my tests to play nicely under python3. Running with nosetest I have 250+ failures, mostly complaining about sqlite. Run each one individually with autotest and they are fine. I need to look into th.is more before I can merge, sorry.

akrabat:
Rebased.

akrabat:
Rebased to use new comparison tests (#854)

.travis.yml
```diff
@@ -13,6 +13,7 @@ language: python
 python:
  - 3.6
  - 3.7
+ - 3.8
```
stephenfin, line 17: You can probably add 3.9 here now, right?

README.rst
```diff
@@ -60,7 +60,8 @@ Install from PyPI
 ~~~~~~~~~~~~~~~~~
 
 The latest released version may be installed from PyPI by using
-``pip``. It supports Python 2.7 or 3.6+::
+``pip``. Version 0.96 was the last version to support Python 2.7.
+Subsequent versions support Python 3.6+::
```
stephenfin, line 64: Maybe say "Python 3 only", so that we don't have to remember to come back and edit this in the future? We're likely documenting our supported Python version elsewhere anyway?

Later: Actually, we don't appear to mention this separately. I still think we should drop this from here since Python support matters whether you're installing from pip or other places. Perhaps we could move this sentence before "Install from PyPI" and rephrase to say e.g. "rst2pdf support Python 3.6 or greater. rst2pdf 0.96 was the last version to support Python 2.7."

README.rst
```diff
@@ -75,8 +76,8 @@ then you can install from there with::
 Install from GitHub
 ~~~~~~~~~~~~~~~~~~~
 
-Work on rst2pdf has restarted on GitHub, with the goals of supporting
-Python 3, addressing outstanding issues, and not breaking anything. You
+Work on rst2pdf has restarted on GitHub, with the goals of adding new
+features, addressing outstanding issues, and not breaking anything. You
 can clone the repository and install this version::
 
     $ git clone https://github.com/rst2pdf/rst2pdf rst2pdf
```
stephenfin, line 83: Unrelated, but you don't need the second argument to `git clone` - it will automatically clone to a directory with the same name as the repo.

rst2pdf/tests/input/sphinx-repeat-table-rows/index.rst
```diff
@@ -6,7 +6,7 @@ This table spans two pages. As the ``pdf_repeat_table_rows`` setting in
 2.
 
 .. table::
-   :widths:: 20 80
+   :widths: 20 80
```
stephenfin, line 9: Wow, how'd I miss this in https://github.com/rst2pdf/rst2pdf/commit/4a795de61b1c3ab93a995c48e28a8ee0e4071e3c

setup.py
```diff
@@ -68,11 +68,9 @@ def read(*rnames):
         'Environment :: Console',
         'License :: OSI Approved :: MIT License',
         'Operating System :: OS Independent',
-        'Programming Language :: Python :: 2',
-        'Programming Language :: Python :: 2.7',
-	'Programming Language :: Python :: 3',
+        'Programming Language :: Python :: 3',
         'Programming Language :: Python :: 3.6',
-	'Programming Language :: Python :: 3.7',
+        'Programming Language :: Python :: 3.7',
```
stephenfin, line 73: ++ I've something similar done in the tox PR

.travis.yml
akrabat: 3.9 isn't [due out until October](https://www.python.org/dev/peps/pep-0596/#schedule).

We should probably add 3.9-dev to Travis at some point. Is Beta1 a good time to do that in a new Python's release cycle?

README.rst
akrabat: Addressed.

README.rst
akrabat: Addressed.

.travis.yml
stephenfin: Ah, I saw it on my PATH and assumed it was the final version. Seems Fedora packages up betas.

We can leave this til release, I suspect, since python's API is pretty stable and we've no history yet of breaking between minor versions

akrabat:
Rebased to latest `master`

setup.py
```diff
@@ -64,18 +77,19 @@ def read(*rnames):
...
+    url="https://rst2pdf.org",
     project_urls={
         'Bug Reports': 'https://github.com/rst2pdf/rst2pdf/issues',
         'Source': 'https://github.com/rst2pdf/rst2pdf',
     },
-    download_url='https://github.com/rst2pdf/rst2pdf/releases',
+    download_url="https://github.com/rst2pdf/rst2pdf/releases",
     entry_points={'console_scripts': ['rst2pdf = rst2pdf.createpdf:main']},
+    test_suite='rst2pdf.tests.test_rst2pdf.test_suite',  # TODO: this needs to be updated
 )
```
stephenfin, line 95: It looks live you've reverted a lot of the changes from c3c8f3dd6 here. Was that intentional?

akrabat: Nope - means I used the wrong command when fixing the conflict!

akrabat:
Fixed rebase error related to setup.py

Issue #601, opened 2016-11-08
AttributeError: PDFWriter instance has no attribute 'custom_cover'


chrchr:
```
Traceback (most recent call last):
  File "/Users/rchurch/sphinx_tst/venv/lib/python2.7/site-packages/rst2pdf/pdfbuilder.py", line 127, in write
    docwriter.write(doctree, destination)
  File "/Users/rchurch/sphinx_tst/venv/lib/python2.7/site-packages/docutils/writers/__init__.py", line 80, in write
    self.translate()
  File "/Users/rchurch/sphinx_tst/venv/lib/python2.7/site-packages/rst2pdf/pdfbuilder.py", line 577, in translate
    log.error("Can't find cover template %s, using default"%self.custom_cover)
AttributeError: PDFWriter instance has no attribute 'custom_cover'
```

This happens with `pdf_cover_template` set in conf.py.

```
$ rst2pdf --version
0.93.dev-r0
$ python --version
Python 2.7.12
$ uname -a
Darwin computer 15.6.0 Darwin Kernel Version 15.6.0: Thu Jun 23 18:25:34 PDT 2016; root:xnu-3248.60.10~1/RELEASE_X86_64 x86_64
```

pshchelo:
To note the bahaviour change - it seems now (with newer Jinja2) the value of `pdf_cover_template` setting should be **short name** (w/o path) to the template instead of a full name.

akrabat:
0.97 is now release with this fix in it.

PR #843, opened 2020-02-10
Template and fonts fixes

pshchelo:
Fixes issue #601 and failures to load fonts


pshchelo:
Updated commit message in the second commit with exact error

akrabat:
Thanks @pshchelo. 

Are you able to create a test that fails without the changes applied and then passes when they are?

pshchelo:
@akrabat  haven't looked at how tests are done in the repo yet, but will try to come up with something

pshchelo:
Came out with a test for #601 , working on font loading one.

@akrabat what is your preferred process re git history? Should I add tests to each corresponding commit, or create a separate commits with tests and place them before the commits with fixes in the PR so that you can checkout only tests commits and see that they are failing? Travis will then check that they are working with fixes..

akrabat:
Separate commits please. Order would be convenient, but not strictly necessary.

pshchelo:
tests added, the one for font loading may not be ideal (as the order of files in os.walk is not guaranteed), but is the best I could quickly think of

pshchelo:
@akrabat plz review

I bet the issue with font loading can affect multiple users.
For example, on my Ubuntu 18.04 system there are several installed by default fonts that font loading fails on (log when running with this patch):
```
[WARNING] findfonts.py:81 Error processing /usr/share/fonts/opentype/noto/NotoSansCJK-Regular.ttc                                                                                                                    
[WARNING] findfonts.py:81 Error processing /usr/share/fonts/opentype/noto/NotoSansCJK-Bold.ttc                                                                                                                       
[WARNING] findfonts.py:81 Error processing /usr/share/fonts/opentype/noto/NotoSerifCJK-Bold.ttc                                                                                                                      
[WARNING] findfonts.py:81 Error processing /usr/share/fonts/opentype/noto/NotoSerifCJK-Regular.ttc                                                                                                                   
[WARNING] findfonts.py:81 Error processing /usr/share/fonts/truetype/noto/NotoColorEmoji.ttf
```
Depending on the order in which Python's `os.walk` finds fonts this can completely break PDF build w/o this patch if such offending fonts are loaded early enough in the list of discovered fonts, producing PDF of zero size.

PR #846, opened 2020-02-21
Change type of pdf_real_footnotes and pdf_use_toc options from str to bool

thijskh:
I added the following to my Sphinx conf.py, conform the rst2pdf documentation:

```py
pdf_use_coverpage = False
pdf_use_toc = False
```

This yields the following output when running sphinx-build:
```
WARNING: The config value `pdf_use_toc' has type `bool', defaults to `str'.
```

So either the type of this variable should be changed, or the documentation updated. I've chosen to change the type.


lornajane:
This causes the sphinx tests to fail, which is they the build is failing. Can you say a bit more about how you use this and show me an example of what this PR helps with? I can take a look at the tests

thijskh:
I guess what I'm trying to do is listed above. I added two said options to my Sphinx config to disable the cover page and table of contents. The options are listed in the manual as booleans, see https://rst2pdf.org/static/manual.pdf page 43.

Building my documentation then generates the warning message above: it laims the setting should be a string while I'm using a boolean in my configuration. Because the documentation says it's a boolean.

To get rid of the warning I propose to change the type of the option to be boolean.

thijskh:
I think what I want to do is similar to 6d76e69c3bc494669b1fdb1b3f60598cb31b71f3

thijskh:
I understand the situation better now and the tests are now passing again.

The original idea of the PR is still the same: the two Sphinx settings `pdf_use_toc` and `pdf_real_footnotes` are documented as boolean, but are actually in the code as strings with defaults `'True'` and `'False'`. This means that if you use them as documented you get warnings and they don't behave exactly as you expect.

This was encountered earlier for another setting and fixed in 6d76e69c3bc494669b1fdb1b3f60598cb31b71f3.

One problem is that the current default values are strings. For `pdf_use_toc` this does not really matter. The old value is string `'True'`, which evaluates to boolean True. but for `pdf_real_footnodes` the current default value is `'False'` which as a string actually evaluates to True. So this means all current installations that do not have an explicit value for `pdf_real_footnotes` have it turned on.

It is clear to me that the intended default for `pdf_real_footnotes` must be `False`, and it's a bug that it defaults to True currently. So I opted to change the default to boolean False. There's one problem though and that's that all sphinx tests were built with this implicit option defaulting to True. Changing the default would make them fail. I opted to solve this to add the previous True explicitly to their configs, so they are running with the settings they expect. This is similar to the approach chosen in 6d76e69c3bc494669b1fdb1b3f60598cb31b71f3.

akrabat:
This'll be a BC break for everyone who doesn't have the `pdf_real_footnotes` setting, but we can code as we're pre 1.0…

akrabat:
Thanks @thijskh

PR #848, opened 2020-03-26
Fix unicode and basestring (required by Python3)

ketoor:


setup.py
```diff
@@ -39,7 +39,7 @@ def read(*rnames):
 setup(
     name="rst2pdf",
     version=version,
-    python_requires='>=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*,!=3.5.*',
+    python_requires='>=2.7,!=3.0.*,!=3.1.*,!=3.2.*',
     packages=find_packages(exclude=['ez_setup', 'examples', 'tests']),
```
oz123: why did you remove python3.5 and python3.4?

ketoor: Sorry, it was intended to cover only single commit

akrabat:
Thanks @ketoor

PR #849, opened 2020-03-27
Switch to image-based comparison

stephenfin:
The checksum-based testing machinery is far too finicky. Checksums change for
all sorts of reasons, many of them environmental. When a checksum does change,
the only way to determine whether it's changed is through manual comparison.
Avoid the whole process and simply do the checksum-based comparison from the
get go. This is slower but, importantly, reproducible.


lornajane:
Ah, this is a great idea, it is certainly true that the tests can be brittle between platforms. However I worry that some of the tests are not purely about presentation, for example some were testing metadata etc, which the checksums also picked up changes in. I'd be concerned about going over to a purely visual approach, but is there a middle ground here?

stephenfin:
> Ah, this is a great idea, it is certainly true that the tests can be brittle between platforms. However I worry that some of the tests are not purely about presentation, for example some were testing metadata etc, which the checksums also picked up changes in. I'd be concerned about going over to a purely visual approach, but is there a middle ground here?

Fair point. What about if we also attempted to compare the output of the [`pdfinfo`](https://linux.die.net/man/1/pdfinfo) on both the reference and generated PDFs? We're already using this to figure out how many pages the document has so we could also use it to inspect other attributes like the title, subject, PDF version, etc. I'd also contemplated extracting all the text from the PDF and comparing that but it seemed like overkill and made the tests slower. It is possible though.

stephenfin:
Consider this blocked by #850 for now too. I think that should make it easier to figure out why this is failing on Travis but not locally.

ralsina:
There are a number of tests where what is being tested is not visible, but something like "where is this linked to" or the structure of the internal PDF table of contents.

In some cases those things could be extracted from the PDF and used as additional sources of comparable data, I guess.

stephenfin:
> There are a number of tests where what is being tested is not visible, but something like "where is this linked to" or the structure of the internal PDF table of contents.
> 
> In some cases those things could be extracted from the PDF and used as additional sources of comparable data, I guess.

Yeah, I've gone away from this and settled on a content-based comparison instead. I have this [ready to go](https://github.com/stephenfin/rst2pdf/tree/content-based-comparison) but am waiting for #850 to merge before I create a PR.

I think I can close this particular PR now.

PR #850, opened 2020-03-31
Switch to pytest

stephenfin:
From the [nose website](https://nose.readthedocs.io/en/latest/index.html):

> Nose has been in maintenance mode for the past several years and will likely cease without a new person/team to take over maintainership. New projects should consider using Nose2, py.test, or just plain unittest/unittest2.

pytest is a good option that gives us much nicer output and makes running individual tests significantly easier. It's also faster. Much faster. On my local machine, current test execution drops by from ~430sec to ~330sec, and if the xdist plugin is enabled this drops even further to ~130sec.


akrabat:
I suspect that this relates to #838.

stephenfin:
> I suspect that this relates to #838.

Damn, didn't realise there was prior art here. However, it looks like #838 is stalled and this is good to go (it's passing CI), so perhaps we should continue with this? Do my changes make sense?

stephenfin:
I do like the approach @normanlorrain used though. I think that's based on [this doc](https://docs.pytest.org/en/2.7.3/example/nonpython.html#yaml-plugin). I could adopt tests to match that in a later approach, if it makes sense?

stephenfin:
I just went and used the approach suggested in #838 since it was cleaner than mine. Hopefully this makes sense now.

akrabat:
I'm speaking at a virtual conference this weekend, so am  not get to it until next weekend.

lornajane:
This is great, I think people really expect to be able to type `pytest` and have the expected thing happen, so I'm very happy to see this. Should we include `pytest` in the requirements file or do you think people will mostly have it installed anyway?

For me, the fancytitles test fails (situation normal) but all my sphinx tests also fail:

```
Running Sphinx v1.8.5

Extension error:
Could not import extension rst2pdf.pdfbuilder (exception: No module named 'rst2pdf')
```

I'm not sure if this is directly linked to this PR though since my python has seemed confused about where `rst2pdf` is when I try to use the local source version of it.

(@akrabat can you poke travis to repeat the build? It succeeded but I don't have rights to make it go again)

stephenfin:
> For me, the fancytitles test fails (situation normal) but all my sphinx tests also fail:
>
> ```
> Running Sphinx v1.8.5
>
> Extension error:
> Could not import extension rst2pdf.pdfbuilder (exception: No module named 'rst2pdf')
> ```
>
> I'm not sure if this is directly linked to this PR though since my python has seemed confused about where rst2pdf is when I try to use the local source version of it.

Ah, these now expect the `rst2pdf` package to be installed itself. Could you try creating a new virtualenv and use the installation instructions from the `.travis.yml` file? That should get things going.

stephenfin:
As for including pytest in the requirements, sure. I only omitted it because nose wasn't included previously. I'll tackle that as a follow up PR though

stephenfin:
> (@akrabat can you poke travis to repeat the build? It succeeded but I don't have rights to make it go again)

I've rebased and force pushed this branch in the hopes of stirring Travis into action. Hopefully that does the trick.

@akrabat Anything else you need from me here?

akrabat:
I had a failure on sphinx-multidoc. Manual validation shows that files produced are correct, so I've added the hash my Mac created.

Good to merge once Travis passes.

PR #851, opened 2020-04-01
Remove dead files

stephenfin:
Remove a number of Python sample files in 'tests/input' that were unreferenced by any test and therefore are surplus to requirements. In addition, remove the 'zinspector12' module, which is unused, rather hacky and easily replaced by an IDE or tools like JEDI.


lornajane:
@akrabat I can't merge since the GitHub doesn't know that the tests pass, can you take a look and merge if you're happy?

PR #855, opened 2020-04-19
tests: Remove noise from Sphinx samples

stephenfin:
- Remove all commented out config options
- Remove everything to do with importing custom extensions manually, which is not something we actually do here
- Remove all config options that match Sphinx defaults or are not used in their respective builds, including: ``templates_path``, ``source_suffix``, ``source_encoding``, ``today``, ``exclude_trees``, ``show_authors``, ``pygments_style``, ``html_*``, ``latex_*``, ``man_*`` and ``htmlhelp_*``
- Fix formatting


akrabat:
I rebased onto the latest master locally and `sphinx-issue168` failed when I ran pytest.

stephenfin:
Oh, I'd forgotten I had done this. I'll rebase this and take a look. I likely removed something I shouldn't have

stephenfin:
Turns out we were including the contents of one of the `conf.py` files in the output. Change the `conf.py` file and you get different output. This is corrected now.

akrabat:
> Turns out we were including the contents of one of the `conf.py` files in the output. Change the `conf.py` file and you get different output. This is corrected now.

For clarity for when we look back at this PR, `sphinx-issue168` has this within its `index.rst`:

```
.. literalinclude:: conf.py
   :lines: 1,3,10-20
```

When Stephen says that this is corrected, he means that the reference PDF has been updated too, to take into account that we've tidied up `conf.py`.

PR #856, opened 2020-04-19
Replace copied file with symlink

stephenfin:
`doc/assets/flowables.py` was an out-of-date copy of `rst2pdf/flowables.py`. Rather than try keep them in sync, simply add a symlink.


akrabat:
Does this have implications for Windows?

stephenfin:
Ah, it would, yes. It looks like symlinks with git require [special configuration on Windows](https://stackoverflow.com/a/40914277/613428). That's unfortunate :disappointed: On the other hand, this would only be an issue if building the documentation on Windows. I assume that's something we want to support?

lornajane:
I think we should probably keep support for building docs on Windows if we can . ... going to close this as I think what we have now is the lesser of two evils!

PR #868, opened 2020-06-05
Stop ignoring generated SVGs

stephenfin:
The presence of these in our input directory indicates a bug in how we're cleaning things up that we should resolve. They shouldn't be ignored.


akrabat:
Agreed. Fortunately, there doesn't appear to be any new tmp*.svg files created when `pyrest` is run at the moment anyway.

Issue #869, opened 2020-06-05
'inkscape' called with invalid argument


stephenfin:
We have [conditionals](https://github.com/rst2pdf/rst2pdf/blob/1ba08758e0c8d3bf78868c313bf21d1132fdb73c/rst2pdf/extensions/inkscape_r2p.py#L62-L64) to handle differences in arguments between Inkscape 1.x and previous versions. Those conditionals state that Inkscape 1.x uses an `--export-file` parameter. However, this is not a valid parameter on the Inkscape 1.0 package on Fedora 31 and looking at the release notes [prove](https://wiki.inkscape.org/wiki/index.php/Release_notes/1.0#Command_Line) that the correct incantation to be `--export-filename`. This [seems to have been a late-breaking change](https://gitlab.com/inkscape/inkscape/-/commit/57793737d53b4187c2474f25f1f6be7a6f118533) that I suspect wasn't present in any of the release candidates, hence why we didn't spot it before.

```
$ inkscape --version
Inkscape 1.0 (4035a4fb49, 2020-05-01)
    Pango version: 1.44
$ sudo dnf info inkscape
Installed Packages
Name         : inkscape
Version      : 1.0
Release      : 2.fc31
Architecture : x86_64
Size         : 188 M
Source       : inkscape-1.0-2.fc31.src.rpm
Repository   : @System
From repo    : updates
Summary      : Vector-based drawing program using SVG
URL          : https://inkscape.org/
License      : GPLv2+ and CC-BY
Description  : Inkscape is a vector graphics editor, with capabilities similar to
             : Illustrator, CorelDraw, or Xara X, using the W3C standard Scalable Vector
             : Graphics (SVG) file format.  It is therefore a very useful tool for web
             : designers and as an interchange format for desktop publishing.
             : 
             : Inkscape supports many advanced SVG features (markers, clones, alpha
             : blending, etc.) and great care is taken in designing a streamlined
             : interface. It is very easy to edit nodes, perform complex path operations,
             : trace bitmaps and much more.
```

akrabat:
Ooh! Good catch. I'm still on a pre-release of 1.0 from January…

PR #870, opened 2020-06-05
Use correct argument to 'inkscape'

stephenfin:
Per #869, inkscape 1.0 final uses '--export-filename' - not '--export-file'. This was causing test failures on Fedora 31.

While we're here, we take the opportunity to clean up this file, replacing some hardcoded paths on Windows and running the whole thing through black. We can do a lot more of the latter as we proceed with fixing various bugs and removing all the skipped tests.

### DIFF
--- a/rst2pdf/extensions/inkscape_r2p.py
+++ b/rst2pdf/extensions/inkscape_r2p.py
@@ -69,11 +69,12 @@ class InkscapeImage(VectorPdf):
                 # Inkscape 0.x uses -A
                 cmd = [progname, os.path.abspath(filename), '-A', pdffname]
             else:
-                # Inkscape 1.x uses --export-file
+                # Inkscape 1.x uses --export-filename
                 cmd = [
                     progname,
                     os.path.abspath(filename),
-                    '--export-file=%s' % pdffname,
+                    '--export-filename',
+                    pdffname,
                 ]
             try:
                 result = subprocess.call(cmd)
@@ -119,11 +120,12 @@ class InkscapeImage(VectorPdf):
                     str(client.def_dpi),
                 ]
             else:
-                # Inkscape 1.x uses --export-file
+                # Inkscape 1.x uses --export-filename
                 cmd = [
                     progname,
                     os.path.abspath(filename),
-                    '--export-file=%s' % pngfname,
+                    '--export-filename',
+                    pngfname,
                     '-d',
                     str(client.def_dpi),
                 ]
--- a/rst2pdf/extensions/inkscape_r2p.py
+++ b/rst2pdf/extensions/inkscape_r2p.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 # See LICENSE.txt for licensing terms
 
-'''
+"""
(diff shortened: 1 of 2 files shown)


rst2pdf/extensions/inkscape_r2p.py
```diff
@@ -58,10 +50,12 @@ def __init__(
         cache = self.source_filecache.setdefault(client, {})
         pdffname = cache.get(filename)
         if pdffname is None:
+            progname = shutil.which('inkscape')
```
akrabat, line 53: Does this work on a stock Windows with Inkscape installed? The [docs](https://docs.python.org/3/library/shutil.html#shutil.which) imply that it searches the path and I'm surprised that `$PROGRAMFILES\Inkscape` is on the path in Windows by default. Does the Windows Inkscape installer add this directory to the path?

i.e. I'd expect this change to mean that it no longer works "out of the box" on Windows.

rst2pdf/extensions/inkscape_r2p.py
```diff
@@ -39,80 +36,118 @@ class InkscapeImage(VectorPdf):
...
 
         pdfuri = uri.replace(filename, pdffname)
         pdfsrc = client, pdfuri
-        VectorPdf.__init__(self, pdfuri, width, height, kind, mask, lazy, pdfsrc)
+        VectorPdf.__init__(
+            self, pdfuri, width, height, kind, mask, lazy, pdfsrc
+        )
+
+    @classmethod
+    def available(self):
```
akrabat, line 91: Which code calls this method?

stephenfin: I've no idea. I simply moved this. We probably want to run things through coverage.py at some point since I'm sure there's plenty of dead code here.

rst2pdf/extensions/inkscape_r2p.py
stephenfin: [This answer on SO](https://stackoverflow.com/questions/56399696/shutil-which-not-finding-programs-without-appending-file-extension) suggests the answer is yes. I don't have a Windows installation to test on though so I think we need to find someone that does else chance it.

rst2pdf/extensions/inkscape_r2p.py
akrabat: Ah - I missed that it was at the top of the file…

rst2pdf/extensions/inkscape_r2p.py
akrabat: I fired up a Windows VM, installed Inkscape and Python and tested. The answer is that the `inkscape.exe` has to be on the path for `shutil.which()` to find it and by default, the Inkscape installer doesn't add its bin directory to the path.

bf614322
 
Incidentally, Inkscape v1's installer places the binary in `C:\Program Files\Inkscape\bin`, so the current default doesn't work going forwards anyway.

akrabat: Essentially, I think we just have to document that Inkscape's binary needs to be on your path.

akrabat:
hmm.. tried to force a re-run of Travis, but it didn't work

PR #871, opened 2020-06-05
Use 'pytest.fail' instead of 'assert' in test

stephenfin:
If we use the latter, we see a really ugly, hard-to-parse traceback. The latter is much more subtle so use that.

The 'pytrace' variable is set to False on all values to avoid triggering a full traceback from 'pytest.fail' also [[1][1]].

[1]: https://docs.pytest.org/en/stable/reference.html#pytest-fail


stephenfin:
We can improve this further by including the output log in the response too. I can do this in a follow up

akrabat:
This is much nicer for the `conftest` errors. A followup that does magic to help with the compare errors would be loverly :)

PR #872, opened 2020-06-05
Cleanup remnants of MD5-based testing suite

stephenfin:
Remove the giant `md5` directory and update some docs to reference the new world.


doc/DEVELOPERS.rst
```diff
@@ -1,13 +1,13 @@
----------------------------
+===========================
```
akrabat, line 1: using `-------` for top level is intentional for this file. See https://github.com/rst2pdf/rst2pdf/commit/6d6490acde1ff05c864936d730d66121e27b3f6c#diff-faab44affa9a75879bed7063892c32f5:

> DEVELOPERS.rst is included in the manual and so it needs to use the same
title levels as manual.rst such that it embeds into the manual below the
chapter level.

i.e. we can't change this without breaking the TOC in the PDF of the manual.

stephenfin: Dropped this commit

PR #875, opened 2020-06-07
Remove Python 2-only code

stephenfin:
This depends on #841. Remove six, unnecessary future imports, and code that is only used under the now-unsupported Python 2.7 environment.

### DIFF
--- a/rst2pdf/basenodehandler.py
+++ b/rst2pdf/basenodehandler.py
@@ -36,7 +36,6 @@ will be logged.
 import inspect
 import types
 
-from six import with_metaclass
 from smartypants import smartypants
 import docutils.nodes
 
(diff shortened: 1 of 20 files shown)


rst2pdf/basenodehandler.py
```diff
@@ -103,7 +102,7 @@ def __init__(cls, name, bases, clsdict):
             cls._classinit()
 
 
-class NodeHandler(with_metaclass(MetaHelper, object)):
+class NodeHandler(metaclass=MetaHelper):
```
stephenfin, line 105: note: Yes, this is correct. You don't need to explicit subclass `object` in Python 3 as discussed [here](https://docs.python.org/3/howto/pyporting.html#subclass-object).

stephenfin:
Resolved the conflicts due to the now-merged #841

Issue #711, opened 2018-11-16
Formatting of the codebase is bad


ralsina:
It's mostly my own fault, too!

But yes, really bad. Now that we have tools like black ... how about a PR that reformats the whole thing?

I can do it if needed.

akrabat:
To be honest, I'd rather we got it running on latest ReportLab and other dependencies with the tests passing first.

Reformatting to a known standard isn't a bad idea, but I really want 0.94 published and working on on a modern Python 2.7.

ralsina:
Ok, can you add a list of things in the 0.94 issue? I can work on it (long weekend!)

oz123:
I also believe that it's better to fix this issue gradually.

oz123:
> Ok, can you add a list of things in the 0.94 issue? I can work on it (long weekend!)

Removing smarty-pants and making sure rson works with python3 would be good.

Also, rson isn't really well documetd, maybe cson could replace it? This would be a breaking change though... https://github.com/avakar/pycson/

PR #877, opened 2020-06-15
Reformat codebase with the Black code formatter

stephenfin:
Now that we've got tests passing consistently on different platforms, it's time to start fixing various bugs and closing feature gaps.
Before we do *that* though, we need to do something about the code style, which is in a somewhat sorry state.
While one should generally avoid applying large, style-only changes to existing code bases because they tend to make a complete mess of our ability to use 'git-blame' and related tools effectively, without this we'd otherwise be fixing this stuff manually for years.
Bite the bullet and run all the code through black [[1][1]] (specifically, `black -S`, to avoid unnecessary conversion of single quotes to double quotes) to address the worst of the issues.
A future change will address the remaining issues like use of wildcard imports and commented out (dead) code.

Note that black defaults to a max line length of 88, not 79 [[2][2]], which means we need to configure flake8 to allow this.

[1]: https://pypi.org/project/black/
[2]: https://black.readthedocs.io/en/stable/the_black_code_style.html

Closes: #724
Closes: #711


akrabat:
Sorry, @stephenfin  - looks like the python3-ification branch conflicted.

stephenfin:
Rebased

akrabat:
Can we add a note to the DEVELOPERs.rst for the exactly `black` command that a contributor should use in order to format their code before raising the PR?

Also, can we validate that the code is formatted correctly using Travis?

akrabat:
My first guess on a travis check is `black --skip-string-normalization --check rst2pdf`

stephenfin:
Working on it

stephenfin:
This is done now. I integrated pre-commit to automate the process of these locally. Let me know if anything doesn't make sense.

PR #881, opened 2020-06-22
Changed `pip` references to `pip3`

nathan-a-macleod:
Pip is for python 2, and pip3 is for python 3 - since python 2 is no longer supported, people need the install links for python 3, not python 2.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -64,7 +64,7 @@ Install from PyPI
 The latest released version may be installed from PyPI by using
 ``pip``.
 
-    $ pip install --user rst2pdf
+    $ pip3 install --user rst2pdf
 
 Install from Snap
 ~~~~~~~~~~~~~~~~~
@@ -84,7 +84,7 @@ can clone the repository and install this version::
     $ git clone https://github.com/rst2pdf/rst2pdf
     $ cd rst2pdf
     $ git checkout <desired-branch> # if you want something other than master
-    $ pip install --user .
+    $ pip3 install --user .
 
 You may want to install it in a virtualenv, but that is beyond the scope
 of this readme.


lornajane:
I'm a little unsure about this because it's platform-specific. I have seen systems that link `pip` to `pip3` (mine does that) and some where `pip` still installs for python2. Is there a safe option (using `pip3` might be it!) that would make this work for most people?

nathan-a-macleod:
I _think_ (if I am interpreting the article correctly) that if Python 2 **isn't** installed of your system but only Python 3 **is**, then `pip` and `pip3` will both install for Python 3 - but if both Python 2 and Python 3 are both installed, then `pip` installs for Python 2, and `pip3` installs for Python 3.

If this is the case, then using `pip3` is probably the better option because it should automatically install for Python version 3 - not version 2.

[This article contains some more information](https://python-forum.io/Thread-What-s-the-difference-between-pip-and-pip3#:~:text=Pip%20is%20for%20python%20version,python%20version%203%20or%20higher.)

akrabat:
Arguably as per the [docs](https://docs.python.org/3/installing/index.html#basic-usage) and to ensure that we work with Windows and with every installation that's [PEP 384](https://www.python.org/dev/peps/pep-0394/) compatible, we should use:

```
python3 -m pip ...
```

nathan-a-macleod:
@akrabat I hadn't seen that before - It that looks like that's the most reliable way to do it for Python 3.

stephenfin:
@nathan-a-mcleod Any chance you could respin this to address @akrabat's comments?

akrabat:
Closing for now. @nathan-a-macleod, if you get time to make the change, please reopen.

Issue #761, opened 2019-03-07
Update pygments stylesheets


ralsina:
We are currently using stylesheets that were generated with the pygments current version of circa 2012. Unsurprisingly, lots have changed in the past 7 years:

* Pygments supports more color schemes
* Pygments stylesheets have more classes
* Some of the class names have changed
* Docutils now uses the long class names for code styling

So, this needs fixing:

* [ ] Update the tool to generate stylesheets from pygments
* [ ] Add the missing color schemes
* [ ] Use (as much as possible) the code highlighter from docutils instead of rst2pdf's own

This is far from trivial, but I sort of know how to do it, so assigning to me.

lornajane:
I'm watching very closely, I'd love to have been able to convert my custom pygments styles over but could never quite figure it out!

ralsina:
@lornajane the current tool kinda sucks but I can do a better one now, since hopefully I am a better programmer than I was in 2009 :-)

akrabat:
I didn't realise there was a current tool!

ralsina:
@akrabat I will add it to the docs somewhere

lornajane:
I'd still love to get the tool and/or help with updating the stylesheets. @ralsina Did you add it to the docs? Do you want to just braindump here and then one of us can run with it?

lornajane:
@ralsina Any more hints on generating pygments stylesheets? I am chomping at the bit for direction here (or if I just should improvise, please say)

lornajane:
We updated to newly generated pygments styles and I can't see us making changes to the syntax highlighting without a much better understand of what problem we are solving and why so I'll close this as I think the main aim was achieved.

PR #882, opened 2020-06-24
Update syntax highlight styles

lornajane:
Looking at #761 and digging around in the codebase, I found `pygments2style.py` which does indeed contain the code that takes in whatever pygments styles are on your computer and turn them into rst2pdf styles. So I've started by dumping the output of that into this branch, it needs more testing and also some investigation into what's now available in pygments that we might not already have support for in rst2pdf.


lornajane:
Updated to say that the newly generated/updated styles do seem to work pretty well. Going to put this PR to not-draft status in case anyone else has an opinion on what else we should do before merging. I'll add a list of styles to the documentation I think as I'm not sure how else they can be discovered.

doc/manual.rst
```diff
@@ -955,20 +955,43 @@ gray. For example, to highlight ``print "line a"`` and ``print "line b"``:
...
 
+* ``abap``
+* ``algol_nu``
+* ``algol``
+* ``arduino``
 * ``autumn``
 * ``borland``
 * ``bw``
 * ``colorful``
+* ``default``
```
akrabat, line 966: Is this style picked up as the default colour scheme if the user doesn't explicitly choose another one?

lornajane:
Rebased to eliminate conflicts from style fixes

doc/manual.rst
lornajane: Yes, this seems to be our default as well as the pygments default style (which is updated by this PR). I get the same output with and without the "default" stylesheet.

lornajane:
Do we need to do anything else before we can merge this styles PR?

PR #887, opened 2020-07-19
Final style fixes

stephenfin:
Finish the great style cleanup of 2020 with a series of fixes intended to get flake8 green. Specifically, we:

- Remove unused imports and variables
- Replace wildcard imports with more specific ones
- Remove code for older versions of Sphinx (<1.0) and ReportLab (<3.0)
- Remove support for wordaxe, which is no longer available on PyPI (this will be replaced by native ReportLab hyphenation support in the future)
- Various others things flake8 complained about

It's probably best to look at individual commits rather than the whole change, since there's quite a lot going on here.


rst2pdf/extensions/fancytitles.py
```diff
@@ -1,14 +1,14 @@
 # -*- coding: utf-8 -*-
+import codecs
+import tempfile
+
```
akrabat, line 4: General question: What's the style rule that determines why this blank line in the imports list is here?

stephenfin: It's import ordering. From PEP8:

> Imports should be grouped in the following order:
>
> Standard library imports.
> Related third party imports.
> Local application/library specific imports.
>
> You should put a blank line between each group of imports.

flake8 won't check these for us since it's out of scope, but other tools like isort can

setup.cfg
```diff
@@ -3,5 +3,5 @@ tag_build = .dev
 
 [flake8]
 max-line-length = 88
-ignore = E501,W503
+ignore = E203,E501,E741,W503
```
akrabat, line 6: I can see why [E741](https://www.flake8rules.com/rules/E741.html) has been added, but why add [E203](https://www.flake8rules.com/rules/E203.html)?

rst2pdf/extensions/fancytitles.py
akrabat: Thanks. Ironically, I came across this just now when looking up up relative imports.

akrabat:
Thanks @stephenfin !

setup.cfg
stephenfin: black was generating code that conflicted with it, and I'd rather disable this and keep black than the other way round

PR #890, opened 2020-07-20
Integrate coverage

stephenfin:
This should help us identify dead code and other optimizations.

### DIFF
--- a/rst2pdf/tests/conftest.py
+++ b/rst2pdf/tests/conftest.py
@@ -43,8 +43,23 @@ def _get_pages(pdf):
 
 
 def compare_pdfs(path_a, path_b):
-    pdf_a = fitz.open(path_a)
-    pdf_b = fitz.open(path_b)
+    try:
+        pdf_a = fitz.open(path_a)
+    except RuntimeError:
+        pytest.fail(
+            'Reference file at %r is not a valid PDF.'
+            % (os.path.relpath(path_a, ROOT_DIR),),
+            pytrace=False,
+        )
+
+    try:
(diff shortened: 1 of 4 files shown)


rst2pdf/tests/conftest.py
```diff
@@ -43,8 +43,23 @@ def _get_pages(pdf):
...
 def compare_pdfs(path_a, path_b):
-    pdf_a = fitz.open(path_a)
-    pdf_b = fitz.open(path_b)
+    try:
+        pdf_a = fitz.open(path_a)
+    except RuntimeError:
+        pytest.fail(
+            'Reference file at %r is not a valid PDF.'
+            % (os.path.relpath(path_a, ROOT_DIR),),
+            pytrace=False,
```
akrabat, line 52: How do I test this? If I remove the Inkscape app from /Application, I get a long traceback starting with `File 'output/test_fancytitles.pdf' was not generated:`. Presumably there's something else I can do that created the PDF, but then causes PyMuPDF to fail?

stephenfin: I was seeing the issue due to a [broken wheel for `PIL`](https://github.com/python-pillow/Pillow/issues/4225) so I'm not sure how you could test with an actual broken package. If you wanted to test this artificially, you could insert some code before this to remove the generated file and replace it with e.g. an empty file or a text file?

akrabat (approved):

akrabat:
It would be nice to know which tests cover which lines of code. Also, we have a lot of tests to write!

akrabat:
Rebased to fix conflict

PR #891, opened 2020-07-20
Improve packaging

stephenfin:
Integrate setuptools-scm to automate versioning based on Git tags, along with other tweaks.


akrabat:
Does this change mean that we need to update `doc/RELEASE_PROCESS.rst` ?

stephenfin:
> Does this change mean that we need to update `doc/RELEASE_PROCESS.rst` ?

Good point. Done.

akrabat (approved):
This is great. Thanks.

PR #896, opened 2020-08-11
Remove dependency on pdfrw

### LINKED ISSUE #895: pdfrw is unmaintained
**Description of problem**

It seems that pdfrw is not maintained: https://github.com/pmaupin/pdfrw/issues/191 and https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=958362. Hence Debian has started to remove it from its repo for the upcoming release. 
The problem is that it means to remove rst2pdf as well and this I would hope it won't happen.

**possible solution?**

The use of pdfrw seems not very wide spread in the rst2pdf codebase. I am not expert of rst2pdf source code but it looks used only here: rst2pdf/extensions/vectorpdf_r2p.py

Do you think the use of pdfrw could be replaced by something else, be it by reimplementing the feature used, importing the needed code, or replace for a maintained dependency?

akrabat:
As per #895, pdfrw is no longer maintained and as a result we risk being pulled from Debian as we're a blocker for [958362](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=958362).

This PR removes our dependency on pdfrw. 

Practically, this means that we no longer support rendering PDFs as images which also means that the option to use Inkscape to render an SVG is removed and [svglib](https://pypi.org/project/svglib/) is the single method we now support for SVG images.

Fixes #895


doc/assets/cover.tmpl
```diff
@@ -1,16 +1,28 @@
+.. class:: logoimage
```
lornajane, line 1: Are we updating the cover due to the inkscape/pdfrw change? Or is this a different thing?

akrabat: The cover had a background image pdf that depended on pdfrw to work. As I had to change that image anyway, I decided to put in the logo and fix the look of the title.

lornajane:
I need to take another look at this, I have the `test_uml_extension` test failing locally (I realise the travis builds pass) and I want to be more sure about why that is happening and what we removed in the tests before I approve.

akrabat:
Would appreciate your eye over this one @stephenfin

akrabat:
Rebased against master after merge of #889

doc/assets/cover.tmpl
```diff
@@ -1,16 +1,28 @@
...
+
+
+.. class:: subtitle
+
+How to use
 
 .. class:: title
 
-{{title}}
+rst2pdf
```
stephenfin, line 18: Was this intentional?

stephenfin (changes requested):
This looks pretty good and seems reasonable. However, you've missed removal of most the reference PDFs for the tests that you've removed, e.g. `rst2pdf/tests/reference/test_issue_127.pdf`, `rst2pdf/tests/reference/test_fancytitles.pdf`, etc. Any reason to keep those or can they go?

Also, as a general point, it'd be nice if the patches were bisectable. In this series, we remove the code dependencies for the `vectorpdf` extension, then remove the extension, and then remove the tests. If we're going to split things up this way (and we probably want to, since the diffs that contain changes to PDFs aren't really readable :smile:) then this should be the other way around.

If you can remove the extra assets, address my inline question and possibly reorder patches to keep things nice and bisectable, I'm +1 :+1:

doc/assets/cover.tmpl
akrabat: Yes, this is intentional. I've split the title on the cover page over two lines with different font sizes.

akrabat:
Referenced files removed. 

I'm not sure what to do about rebased to reorder as it won't reflect my thinking when I did the work. Maybe that doesn't matter? However, to do the work, I essentially broke it and then fixed it until it worked again :)

stephenfin:
> Referenced files removed.

Looks like you missed `rst2pdf/tests/reference/sphinx-issue251.pdf`. Can you remove that too?

> I'm not sure what to do about rebased to reorder as it won't reflect my thinking when I did the work. Maybe that doesn't matter? However, to do the work, I essentially broke it and then fixed it until it worked again :)

That's how I would have approached it too :smile: After doing so though, I'd personally rewrite history to build a more meaningful "story" for reviewers to follow. As things stand, the series is kind of like deciding to tear down a house and starting with the foundations. The reworked history is a nice-to-have though so nothing I'd hold stuff up for.

rst2pdf/tests/input/test_issue_239_2.txt
```diff
@@ -1,3 +0,0 @@
-.. |warning| image:: warning.svg
```
lornajane, line 1: Did we lose all SVG support with losing inkscape?

rst2pdf/tests/input/test_issue_266.txt
```diff
@@ -1,34 +0,0 @@
...
-
-Enlarged PDF image:
-
-.. image:: images/text_and_vectors_001.pdf
-   :scale: 200%
-
-
-Normal image:
-
-.. image:: images/middleware.jpg
```
lornajane, line 23: Is there value in keeping a test with rescaled JPGs? I realise the PDFs won't work

rst2pdf/tests/input/test_issue_267.txt
```diff
@@ -1,90 +0,0 @@
...
-   :align: right
-   :figwidth: 50%
-
-Align center:
-
-.. figure:: images/biohazard.png 
-   :align: center
-   :figwidth: 50%
-
-Alignment for PDF-based images:
```
lornajane, line 46: Could we keep the test with the png images and just remove the PDFs and update the test?

lornajane (commented):
I think this is sound and a really sound response to not having a pdfrw library. I had a couple of questions and if you want to drop the test-rescue into a set of issues and deal with them outside this PR, that would be fine too. Let's review if we want to update tests now or later, and then review again

akrabat:
> > Referenced files removed.
> 
> Looks like you missed `rst2pdf/tests/reference/sphinx-issue251.pdf`. Can you remove that too?
> 

`sphinx-issue251` was modified to remove the declaration of the vectorpdf extension, not removed. Mind you, it's actually ignore, so needs investigation at some point!

rst2pdf/tests/input/test_issue_239_2.txt
akrabat: No. SVG support remains supported vis [Svglib](https://pypi.org/project/svglib/).

rst2pdf/tests/input/test_issue_266.txt
akrabat: Very good point. I've restored as `test_jpg_scaling`.

rst2pdf/tests/input/test_issue_267.txt
akrabat: Yes. I've restored as `test_alignment_of_figures_and_images`.

akrabat:
Address your item @stephenfin & I've re-added to two test files @lornajane. Would appreciate it if you could give it another look.

stephenfin (approved):

akrabat:
Rebased to fix conflicts.

rst2pdf/tests/input/test_issue_239_2.txt
bmxp: But svglib will not provide support for gradients ...

PR #901, opened 2020-08-23
Rename plantuml extension so it loads before the generic pip plantuml

lornajane:
I've had the `test_uml_extension` test failing for a while - I'm never really sure if it's a problem on my system, a bug in the branch I'm testing or what. Today we (I had help from @akrabat ) identified that if you do `pip install plantuml` then you get the errors I was seeing - and that it will load that extension instead of the one in the `extensions` directory.

However if there's a file called `[module_name]_r2p.py` in `/extensions` then it'll load that and then load the pip module. We should look out for other naming clashes, but this change solves my particular failing test (I assume other people/travis) didn't have the plantuml module from pip installed.


rst2pdf/extensions/plantuml_r2p.py
```diff
@@ -92,5 +92,4 @@ def gather_elements(self, client, node, style):
         # Add Image node with the right image
         return [MyImage(tfile.name, client=client)]
 
-
```
stephenfin, line 95: This was accidental, I assume?

stephenfin (commented):
Makes sense. Looks like we append the `_r2p` suffix when trying to find extensions (see `rst2pdf.createpdf.add_extensions`) so this makes sense to me.

Just need to re-add that dropped line and we're golden.

stephenfin (commented):
Makes sense. Looks like we append the `_r2p` suffix when trying to find extensions (see `rst2pdf.createpdf.add_extensions`) so this makes sense to me.

Just need to re-add that dropped line and we're golden.

stephenfin (changes requested):
Makes sense. Looks like we append the `_r2p` suffix when trying to find extensions (see `rst2pdf.createpdf.add_extensions`) so this makes sense to me.

Just need to re-add that dropped line and we're golden.

stephenfin (changes requested):
.

rst2pdf/extensions/plantuml_r2p.py
lornajane: I think it just looked wrong to me, I thought I was tidying up! I can remove it if it's going to cause a problem

stephenfin: Yeah, the `lint` job will fail with this because it's not PEP-8 compliant. You need to re-add this for it to merge.

lornajane: got it! I had a lot of debug in there and didn't realise I hadn't quite reset it all. The build did fail so it should have been obvious the PR needed work.

stephenfin (approved):

PR #904, opened 2020-08-27
Fix docutils class names in manual

nfraprado:
The docutils classes item-list, bullet-list, option-list, field-list,
item-list-item and bullet-list-item are referred in the manual with
underlines instead of dashes, making them invalid docutils class names.
Fix those references in the manual.

### DIFF
--- a/doc/manual.rst
+++ b/doc/manual.rst
@@ -726,7 +726,7 @@ applied to a table.  This happens in two cases:
    table. Currently these include:
 
    * Footnotes / endnotes (endnote style)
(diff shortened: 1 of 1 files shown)


stephenfin (approved):
LGTM

akrabat (approved):

q2dg:
Where is a comprehensive list of default available classes ready to use? I've looked at "styles.style" file but I hope there is another more complete and descriptive reference source Thanks!

akrabat:
@q2dg There isn't a list unfortunately. Ideally, we'd have one in the manual.

PR #908, opened 2020-09-01
Escape the regex to remove a FutureWarning. Fixes #880

### LINKED ISSUE #880: Possible nested set FutureWarning with Python 3.8
When running `rst2pdf` with Python 3.8 (and, I suspect, 3.7 too), you see the following warning:

```
/usr/lib/python3/dist-packages/rst2pdf/rson.py:140: FutureWarning: Possible nested set at position 2
  splitter = re.compile(pattern).split
```

(this is a Debian unstable box, with version 0.97-2 of rst2pdf but I just checked and the `rson.py` in the repo on the head of master seems the same)

With a bit of digging, it turns out that this is Python warning you that they might add support for nested sets in regexes in the future. Something like this: `r'foo[[ba]r]`, I think. Search for "nested sets" in the [What's new in Python 3.7](https://docs.python.org/dev/whatsnew/3.7.html) announcement for details.

Anyway, I think that this is caused by the unescaped `[` that comes from `delimiterset` (through `re_delimiterset`, `delimiter_pattern` and ends up in `pattern`). I suspect that all you need is to add an escape character to `[` as well as `]` in the definition of `re_delimiterset`.

🖥 **Versions**
`python3 -V` gives: `Python 3.8.3`
`pip freeze | grep rst2pdf` gives `rst2pdf==0.97.dev0`
`pip freeze | grep reportlab` gives `reportlab==3.5.34`

This is a Debian unstable box.

lornajane:
Thanks to a really excellent bug report (thanks @rswarbrick), I was able to look at the regex that triggers a FutureWarning in `rson.py` (line 119, I'm so tired of this warning!) and add a little extra escaping. I can't replicate the same warning message with this change.

### DIFF
--- a/rst2pdf/rson.py
+++ b/rst2pdf/rson.py
@@ -88,7 +88,7 @@ class Tokenizer(list):
     # RSON syntax delimiters are tokenized separately from everything else.
     delimiterset = set(' { } [ ] : = , '.split())
 
-    re_delimiterset = ''.join(delimiterset).replace(']', r'\]')
+    re_delimiterset = ''.join(delimiterset).replace(']', r'\]').replace('[', r'\[')
 
     # Create a RE pattern for the delimiters
     delimiter_pattern = '[%s]' % re_delimiterset


rst2pdf/rson.py
```diff
@@ -88,7 +88,7 @@ class Tokenizer(list):
     # RSON syntax delimiters are tokenized separately from everything else.
     delimiterset = set(' { } [ ] : = , '.split())
 
-    re_delimiterset = ''.join(delimiterset).replace(']', r'\]')
+    re_delimiterset = ''.join(delimiterset).replace(']', r'\]').replace('[', r'\[')
```
rswarbrick, line 91: A "less magic" version of this would be something like:
```python3
delimiterset = set(' { } [ ] : = , '.split())
delimiter_pattern = '[{}]'.format(''.join(re.escape(c) for c in delimiterset))
```
You end up with:
```
>>> delimiter_pattern
'[:\\]\\[\\{\\},=]'
```

rswarbrick: (of course, this only works if no-one is using `re_delimiterset`)

akrabat:
Interestingly, the failed CI is a result of https://github.com/rst2pdf/rst2pdf/pull/855. What I can't explain is why any CI since that was merged passed!

rst2pdf/rson.py
lornajane: I'm not Python enough to understand the "of course" remark here. Is there a danger this change could break something for someone? I'd be happy to adopt the "less magic" version if "less magic" means "more recommended"

lornajane:
Rebased to take in the test/build fixes that are now on master

rst2pdf/rson.py
stephenfin: I guess he simply means if no-one is relying on the variable being present. What's suggested there is indeed better and would be a great follow up. What you have is perfectly suitable for now though.

stephenfin (approved):

rst2pdf/rson.py
rswarbrick: @stephenfin Yep, that's what I meant. Sorry for the slow reply.

Issue #910, opened 2020-09-02
Test pass even when PDF's are different


akrabat:
The tests can pass even if the PDFs are different.

I've been looking at `pytest input/sphinx-issue183`. On the cover page, the date is `April 29, 2018`, which is hard coded via the `today` setting in `input/sphinx-issue183/conf.py`.

If I set `today` to a variety of different dates, sometimes the test passes even though the date on the cover page is not the correct date.

Tests:

* `today = 'August 28, 2020'` => Test Passes
* `today = 'May 28, 2020'` => Test Passes
* `today = 'January 15, 1488'` => Test Passes
* `today = 'September 28, 2020'` => Test Fails

lornajane:
I don't have a `today` setting in my `conf.py` for this test, I think the date is always the current date when I run the tests

akrabat:
> I don't have a `today` setting in my `conf.py` for this test, I think the date is always the current date when I run the tests

That's correct. There was a `today` in the `conf.py` until we merged #855 which removed it by accident. That PR passed the tests though, so no-one spotted it.

PR #909 restores the `today` setting to the affected tests, but we really need the tests to fail reliably when something like this happens.

PR #911, opened 2020-09-02
Ensure fontList is set for a previously embedded font

### LINKED ISSUE #902: fontName in style not working with fontsAlias
It seems that specifying a font as `stdFont` in `fontsAlias`, then later using `fontName: stdFont` in some style doesn't work.
Instead of using the alias it seems to be taking `stdFont` as a literal font name.

**Description of problem**

1. Download the rst and style files for "Document with Simple Styling" from the [rst2pdf examples page](https://rst2pdf.org/examples).
2. Generate the pdf
3. Change the `fontName` for the `normal` style to be literally `Georgia`
4. Regenerate the pdf
5. Notice that the font changed, even though `stdFont` should've been pointing to the `Georgia` font.

**What is the expected output? What do you see instead?**
Using `stdFont`, which is `Georgia`, or using directly `Georgia` should both yield the same pdf.

🖥 **Versions**

`python -V`: Python 3.8.5

`pip freeze | grep rst2pdf`: rst2pdf==0.97.dev0

`pip freeze | grep reportlab`: reportlab==3.5.48

**Which operating system are you using?**
Archlinux 5.8.3-arch1-1

**Additional information**

I don't actually have the Georgia font, but using `pdffonts document1-simple.pdf` (at step 2.) I can see that the font used was `DejaVuSans`, even though `fc-match Georgia` yields `DejaVu Serif`.
Using `pdffonts document1-simple.pdf` again at step 4. shows `DejaVuSerif`.
Also, `fc-match stdFont` returns `DejaVu Sans`, so that's why I think `stdFont` is being taken literally as the font name.

akrabat:
If a font is aliased and then that font name is subsequently used as a style name, then the `fontList` variable will not be set which will result in the font name becoming an alias to whichever font happened to be in `fontList` the previous time around the loop.

Example style sheet:

```
fontsAlias:
    stdFont: inconsolata
    stdBold: carlito

styles:
    normal:
        fontName: inconsolata
        textColor: #383745
        fontSize: 16
        alignment: TA_LEFT

    heading:
        fontName: stdFont
        fontSize: 32
        textColor: #1a8099
```

In this example, `inconsolata` is set as an alias to `stdFont` which means that `inconsolata` is registed as the key name for the embedded font list pointing to the correct font file names for that font (`Inconsolata-Regular_9ba2497f55c3435c544f6702d57e7a94` on my machine). However, `inconsolata` is also referenced as the `fontName` for the `normal` style which means that without this fix, the embedded font referenced by the `inconsolata` key now points the last font that was processed (`Cardo` in this case).

Fixes #902.

Note: The CI will fail until #909 is merged (and ideally #910 is fixed). I will then rebase.


nfraprado:
I tested the change in `rst2pdf/styles.py` and it works as expected, fixing the bug.

akrabat:
Rebased against master now that #909 has been merged.

rst2pdf/styles.py
```diff
@@ -363,7 +363,7 @@ def __init__(self, flist, font_path=None, style_path=None, def_dpi=300):
                         fname, pos = findfonts.guessFont(style[key])
 
                         if style[key] in embedded_fontnames:
-                            pass
+                            fontList = findfonts.autoEmbed(style[key])
                         else:
                             fontList = findfonts.autoEmbed(style[key])
```
stephenfin, line 368: Seeing as these are now identical, we could move them out of the conditional are simply retain the rest of the else block.

```
fontList = findfonts.autoEmbed(style[key])
if style[key] not in embedded_fontnames:
    if fontlist:
        # ...
```

stephenfin (approved):
This looks sensible. One comment inline but nothing I'd hold this up over

stephenfin (approved):
Still good

lornajane (approved):
Yep, works for me. Thanks for fixing :+1:

PR #913, opened 2020-09-07
Document fontSize usage with percentage values

nfraprado:
The fontSize attribute also permits percentage values which are relative
to the style's parent, but this behavior wasn't yet documented.

Add paragraph and example describing its usage in the "Style definition"
section of the manual.


stephenfin (approved):

lornajane:
Thank for spotting this and updating the manual, that's very helpful and much appreciated :+1:

Issue #912, opened 2020-09-04
Generated manuals for rst2pdf.github.io have version labelling issues


lornajane:
**Description of problem**

I'm reviewing the pull request to update the manuals for v0.98 on the docs repo: https://github.com/rst2pdf/rst2pdf.github.io/pull/20/files and I am seeing some small problems with the versions.

On the PDF version of the manual:
 - the coverpage shows the version as `0.98 (dev119)`

On the HTML version of the manual:
 - the metadata looks like it is missing some version information (this relates to the changes made since the 0.97 release in #891)
 - take a look at [this section of the diff](https://github.com/rst2pdf/rst2pdf.github.io/pull/20/files#diff-b378d6f8144070ae35e6265270ec5a45L9-L10) where the previous output had version numbers.

(opening the issue on this repo as it's this one that needs changes!)

stephenfin:
Sigh, this is me too. Apologies. Hopefully I'll get to this this weekend.

lornajane:
Don't apologise! You are making such a positive impact here

PR #915, opened 2020-09-08
docs: Drop revision, version information from meta

stephenfin:
The 'meta' directive is a HTML-builder specific directive used for
generating '<meta>' tags in the output [1]. It is also supported by
Sphinx [2].

In #891, we added support for two new roles, 'package-version' and
'package-revision', that would allow us to programmatically include the
version and revision for a package in documentation. These could be used
to programmatically set the values for '<meta name="version">' and
'<meta name="revision">' tag...or so we thought. Unfortunately the
directive must contain a *flat* field list [1], meaning it's not
possible to include roles inside a field list and the roles are simply
rendered literally.

There are couple of solutions: we could go back to setting the package
version manually in the documentation, we could generate a separate file
as part of our build process and include that in the manual, or we could
simply remove these tags from the docs in favour of explicit inline
docs. We choose the final option since meta tags aren't user-visible (at
least not without inspecting the HTML) and therefore aren't massively
important.

[1] https://docutils.sourceforge.io/docs/ref/rst/directives.html#meta
[2] https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#html-metadata

Closes: #912

### DIFF
--- a/doc/manual.rst
+++ b/doc/manual.rst
@@ -3,9 +3,7 @@ How to use rst2pdf
 ==================
 
 .. meta::
-  :authors: rst2pdf project <https://rst2pdf.org>; Roberto Alsina <ralsina@netmanagers.com.ar>  and the contributors to the rst2pdf project;
-  :version: :package-version:`rst2pdf`
-  :revision: :package-revision:`rst2pdf`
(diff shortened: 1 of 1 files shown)


lornajane (approved):
This is a sensible fix, thanks!

PR #916, opened 2020-09-08
Fix typo

stephenfin:
We were inadvertently comparing a string against itself. It seems we
need tests for our tests /o\

Closes: #910


akrabat (approved):
Doh! That's an easy one to miss!

akrabat:
Thanks @stephenfin

PR #925, opened 2020-09-25
Set return code on error

### LINKED ISSUE #886: rst2pdf should return non-zero on failure
The rst2pdf command line tool should return zero on success and non-zero on failure. This would help scripts that integrate with it and make it a better citizen of the terminal.

akrabat:
When rst2pdf errors, sometimes it returns a non-zero code, but often it doesn't, even though you can see an "ERROR" line in the output. These seem to be from docutils where by default an ERROR isn't considered important enough to stop processing.

This PR does two things:

* Ensures we return 1 on some known errors
* Configures docutils to error with a non-zero return code when it encounters an error

I'm comfortable with this as I think that if docutils has reported an error then you really should fix it as the output of your PDF is almost certainly not what was intended.

To do:

* Check what if any errors can occur after docutils parsing and return non-zero for those cases too.

Fixes #886


stephenfin:
:partying_face:

CONTRIBUTING.rst
```diff
@@ -47,6 +47,7 @@ Tests are in ``rst2pdf/tests`` directory. Within this folder:
   - ``test_[name].txt`` holds the rst source for the test (yes, really)
   - ``test_[name].style`` has any stylesheet that should be applied for this test
   - ``test_[name].cli`` allows adding extra commands to the test (look for examples)
+  - ``test_[name].exitcode`` allows for specifying an expected non-zero exit status code for expected error cases
```
stephenfin, line 50: It's a nit, but can we call this retcode? That seems to be more common

rst2pdf/tests/conftest.py
```diff
@@ -216,7 +216,21 @@ def runtest(self):
         # verify results
 
         if retcode:
-            self._fail('Call failed with %d' % retcode, output)
+            exitcode_file = os.path.join(INPUT_DIR, self.name + '.exitcode')
```
stephenfin, line 219: This will catch cases where it returns a non-zero error code when it shouldn't, but not the other way around. We should check for the exitcode (retcode) file regardless.

rst2pdf/tests/conftest.py
```diff
@@ -309,27 +323,29 @@ def _build(self):
...
+                    'File %r was not generated'
+                    % (os.path.relpath(output_file, ROOT_DIR),),
+                    output,
+                )
+            elif os.path.exists(output_file) and no_pdf:
+                self._fail(
+                    'File %r was erroneously generated'
+                    % (os.path.relpath(output_file, ROOT_DIR),),
+                    output,
+                )
```
stephenfin, line 340: Why the move? This will mean we won't get a log if either of these conditions occurs

stephenfin (changes requested):

nfraprado:
This PR should have a "Fix #886", so that that issue links to this PR.

Also, I just made a PR #974 that will cause conflicts with the `CONTRIBUTING.rst` change in this one, but since the change in this PR already needed a rebase, I figured it was ok.

CONTRIBUTING.rst
akrabat: Done.

rst2pdf/tests/conftest.py
akrabat: Done.

rst2pdf/tests/conftest.py
akrabat: I missed what `.nopdf` does. I've put this code block back and documented `.nopdf`…

akrabat:
> This PR should have a "Fix #886", so that that issue links to this PR.

So it should! Added.

> Also, I just made a PR #974 that will cause conflicts with the `CONTRIBUTING.rst` change in this one, but since the change in this PR already needed a rebase, I figured it was ok.

I've rebased against latest master.

akrabat:
Marked as Ready for Review as it's better to get this in and then iterate than wait for perfection

rst2pdf/tests/input/test_issue_274_2.rst
```diff
@@ -171,7 +171,7 @@ ffff
 * fff f fçff föfffşff fffıffıffffı füfüfff.
 * fff ff fçff ffffff fffıfı üfffffff çffışıfıfff.
 * b f İş ffffffff, c f ffffff füfffffff ffıfı ffffffffffı.
-* İş fffıfffff fffüfffı'fff fffff_ ff fe-fff_ fçff fıfff fffffüf fff.
+* İş fffıfffff fffüfffı'fff fffff ff fe-fff_ fçff fıfff fffffüf fff.
```
nfraprado, line 174: I understand that the `_` at the end of the word here creates a reference and therefore was removed, but there is another one on this same line, and also another on line 262. Shouldn't they be removed as well?

Also, why was the reference pdf removed? Won't that cause the test to fail since it can't be compared to the reference?

I think you should removed these `_` and add the updated reference pdf.

lornajane: I think this was removed because it was causing a docutils error, that is normally just logged and we continue. Now that we've set rst2pdf to error if there is a docutils error, we need to remove this error from the test! @akrabat does that sound plausible?

lornajane (commented):
This is in good shape although I'm sure I have rst2pdf documents that emit errors and still work, I don't have a replication case so let's proceed. I would like to see the small amendments to the contribution file and also I'm sorry but I think this branch needs to rebase again to pick up the build fix we put in master a few days ago

rst2pdf/tests/input/test_issue_274_2.rst
akrabat: > I understand that the `_` at the end of the word here creates a reference and therefore was removed, but there is another one on this same line, and also another on line 262. Shouldn't they be removed as well?

Docutils doesn't detect the other two as references so they don't cause errors. I think this is because a simple reference is a trailing underscore on a set of ascii characters, so `fe-fff_` and `föfffföf_` do not qualify, but `fffff_` does.

> Also, why was the reference pdf removed? Won't that cause the test to fail since it can't be compared to the reference?

The reference PDF (rst2pdf/tests/reference/test_issue_274_2.pdf) hasn't been removed.

CONTRIBUTING.rst
```diff
@@ -51,6 +51,9 @@ Tests are in ``rst2pdf/tests`` directory. Within this folder:
     case they aren't met, the test is skipped. Each dependency should be added
     in a separate line. To add a new one, first add it in the
     ``check_dependency`` list inside ``conftest.py``.
+  - ``test_[name].retcode`` allows for specifying an expected non-zero return code for expected error cases
+  - ``test_[name].retcode`` allows for specifying an expected non-zero return code for expected error cases
```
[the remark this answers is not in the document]
akrabat: Double the benefit! (Fixed!)

CONTRIBUTING.rst
```diff
@@ -51,6 +51,9 @@ Tests are in ``rst2pdf/tests`` directory. Within this folder:
...
+  - ``test_[name].nopdf`` is an text file whose existence indicates that the test does not create a PDF output file
```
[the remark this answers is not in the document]
akrabat: Good catch. Fixed. Thanks.

akrabat:
Rebased and addressed issues in `CONTRIBUTING.rst`.

lornajane:
the github diff view does show that we have an empty PDF file - but the actual diff (meaning I have both branches on my actual laptop) shows some minor changes but the new PDF is essentially alive and well - I didn't understand this question from @nfraprado but now I am glad it made me take a second look!

lornajane (approved):
Seems sensible and I haven't managed to break it so ... approved!!

Issue #923, opened 2020-09-25
Rename test files to `.rst` 


akrabat:
Why do all our input test rst files have `.txt` extensions? `.rst` would make it easier for editors to syntax highlight them…

stephenfin:
This is a [docutils recommendation](https://docutils.sourceforge.io/FAQ.html#what-s-the-standard-filename-extension-for-a-restructuredtext-file) that I happen to disagree with. I agree that changing things would make sense.

akrabat:
> This is a [docutils recommendation](https://docutils.sourceforge.io/FAQ.html#what-s-the-standard-filename-extension-for-a-restructuredtext-file) that I happen to disagree with. I agree that changing things would make sense.

Good grief!

PR #929, opened 2020-10-04
Renaming .txt files to .rst in the '/tests/input' directory

diegofsousa:
Hello everyone, everything good? This is my first contribution to this repository based on issues.

This PR intends to resolve issue #923.

All ".txt" files for input testing have been renamed to ".rst"

PR #930, opened 2020-10-04
Rename txt test files to rst. #923

JulianPedro:
Only files starting with test_ had their txt extensions for rst


JulianPedro:
Only files starting with test_ had their txt extensions for rst

lornajane:
Thanks for this! I have to point out that we had another pull request for the same thing opened immediately before this one  (see #929 ) however that one and this one both stopped most of our tests from running. There should be 297 tests, but as you can see from the build output, only 32 of them run when this change is made. I'll close this PR as a duplicate but I did want to thank you and offer feedback on the PR.

PR #933, opened 2020-10-14
doc: Document hyphenation using pyphen

### LINKED ISSUE #888: Update manual to explain use of pyphen
ReportLab supports hyphenation natively now using pyphen library along with the style settings `embeddedHyphenation` and `hyphenationLang`.

We need to document this in the manual.

These are the steps I did to test it:

Firstly, I installed pyphen with: `pip3 install pyphen`

Then:

test_hyphenation.rst:

```rst
Hyphenation Test
================

You are hereby charged that on the 28th day of May, 1970,
you did willfully, unlawfully, and with malice of forethought,
publish an alleged English-Hungarian phrase book with intent
to cause a breach of the peace. How do you plead?
```

test_hyphenation.style:

```
pageSetup:
    firstTemplate: twoColumn
    margin-left: 2cm
    margin-right: 2cm

styles:
    base:
        fontSize: 13
        leading: 15
        embeddedHyphenation: 1
        hyphenationLang: en_GB
```

Run: `rst2pdf -s test_hyphenation.style test_hyphenation.rst` to get a PDF file containing:

Screenshot 2020-07-19 at 10 51 11

nfraprado:
Update the manual's Hyphenation chapter to use pyphen instead of the deprecated wordaxe. Fixes #888.

From my tests, the `-l` flag for rst2pdf didn't work for hyphenation contrarily to what the `--help` says, so I dropped the documentation text regarding that flag.
If that flag will only support setting the docutils language moving forward, I could also send a PR fixing the `--help` text for `-l`.

I also dropped the hyphenation preview for the spanish text, since it already wasn't working and I didn't see much other use of `.. class` in the manual. To get this to work would probably involve adding a style with the spanish language and lower text width and using its class here. If you think it would be a good addition I can add it to this PR.

### DIFF
--- a/doc/manual.rst
+++ b/doc/manual.rst
@@ -1510,53 +1510,35 @@ Hyphenation
 
 If you want good looking documents, you want to enable hyphenation.
 
-To do it, you need to install Wordaxe [#]_.
+To do it, you first need to install the ``pyphen`` python module.
 
-.. [#] Use Roberto Alsina's fork of Wordaxe from https://github.com/ralsina/wordaxe as
-       this works with later versions of ReportLab.
+Then, you need to specify the language in each style that you want hyphenation
+to work. To have hyphenation in the whole document, you can do it in the
+``base`` style.
 
-If after installing it you get the letter "s" or a black square instead of a
-hyphen, that means you need to replace the rl_codecs.py file from reportlab with
-the one from wordaxe.
+For example, for an English document, hyphenation can be turned on for the whole
+document with::
 
-For more information, see `this issue`_ in rst2pdf's bug tracker.
-
-.. _this issue: http://code.google.com/p/rst2pdf/issues/detail?id=5
-
-Also, you may need to set hyphenation to true in one or more styles, and the
-language for hyphenation via the command line or paragraph styles.
-
-For English, this should be enough::
-
-    ["bodytext" , {
-      "alignment": "justify",
-      "hyphenation": true
+    ["base" , {
+      "hyphenationLang": "en_US",
+      "embeddedHyphenation": 1
     }],
 
-If you are not an English speaker, you need to change the language using
-the ``-l`` or ``--language`` option.
-
-Since Wordaxe version 0.2.6, it can use the PyHyphen library if it's
-available. PyHyphen can use any OpenOffice dictionary, and can even download
-them automatically. [#]_
-
-.. [#] For more information, please check the PyHyphen website at http://pyhyphen.googlecode.com
-
-For example, this will enable German hyphenation globally::
-
-    rst2pdf -l de_DE mydocument.txt
+Notice the ``embeddedHyphenation`` option. It is optional, but it makes so that
+hyphenations will give preference to splitting words at embedded hyphens in the
+text.
 
 If you are creating a multilingual document, you can declare styles with
 specific languages.  For example, you could inherit ``bodytext`` for Spanish::
 
     ["bodytext_es" , {
       "parent": "bodytext",
-      "alignment": "justify",
-      "hyphenation": true,
-      "language": "es_ES"
+      "hyphenationLang": "es_ES",
+      "embeddedHyphenation": 1
     }],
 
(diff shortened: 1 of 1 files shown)


doc/manual.rst
```diff
@@ -1502,53 +1502,35 @@ Hyphenation
...
-.. [#] Use Roberto Alsina's fork of Wordaxe from https://github.com/ralsina/wordaxe as
-       this works with later versions of ReportLab.
+Then, you need to specify the language in each style that you want hyphenation
+to work. To have hyphenation in the whole document, you can do it in the
+``base`` style.
 
-If after installing it you get the letter "s" or a black square instead of a
-hyphen, that means you need to replace the rl_codecs.py file from reportlab with
-the one from wordaxe.
+For example, for an english document, hyphenation can be turned on for the whole
```
akrabat, line 1511: Capital _E_ in _English_ please.

doc/manual.rst
```diff
@@ -1502,53 +1502,35 @@ Hyphenation
...
-available. PyHyphen can use any OpenOffice dictionary, and can even download
-them automatically. [#]_
-
-.. [#] For more information, please check the PyHyphen website at http://pyhyphen.googlecode.com
-
-For example, this will enable German hyphenation globally::
-
-    rst2pdf -l de_DE mydocument.txt
+Notice the ``embeddedHyphenation`` option. It is optional, but it makes so that
+hyphenations will give preference to splitting words in embedded hyphens in the
```
akrabat, line 1520: Please change the _in_ in "words in embedded hyphens" to _with_, so that it reads "words with embedded hyphens".

akrabat (changes requested):
This is great & much more useful to someone who wants hyphenation in their documents. Just a couple of changes as noted please.

akrabat:
I wonder if we should set `embeddedHyphenation` to `1` in the default stylesheet and also if we should set `hyphenationLang` to whatever `--language` has been set to for `base` too?

doc/manual.rst
nfraprado: I guess what I meant was actually `at` rather than `in`. So, "splitting words at embedded hyphens". Because from my understanding that option tries to split the word at the embedded hyphen if it is present.
Which do you think is the best?

nfraprado:
Default `embeddedHyphenation` sounds good. I'd think it adds an unnoticeable overhead for a feature that is almost always useful.

Setting `hyphenationLang` for `base` from `--language` also sounds useful. I could see someone using the same stylesheet for documents in different languages and specifying them in the command line rather than inside the stylesheet.
We would just need to clarify behavior. I'd think that if `base` already had `hyphenationLang`, and `--lang` was passed, `--lang` would have precedence and would override the stylesheet. However, since this would be done only for `base`, sub-styles could specify their own language without ever being overridden by `--lang` (like the `bodytext-es` given as example in the manual).
If we do this I should probably bring back the last paragraph and tip that I removed from the manual, updating it to match whatever we do here.

doc/manual.rst
akrabat: `at` works better for me.

nfraprado:
> I wonder if we should set `embeddedHyphenation` to `1` in the default stylesheet and also if we should set `hyphenationLang` to whatever `--language` has been set to for `base` too?

I was looking into this and realized we also need to be able to disable hyphenation even if it inherits `hyphenationLang` from its parent. From testing, the only way I could do this was by setting it to `0`. Both setting it to `None` or to arbitrary values caused errors.

I added a paragraph documenting how to disable hyphenation.

lornajane:
Thanks so much for this! The build failed for other reasons, which are now fixed (thanks @akrabat and PR #940 ). Could you therefore rebase this PR onto the master branch to pick up those changes and hopefully the build will behave better this time.

akrabat:
I've moved the discussion of automatically enabling this to #942. So we can concentrate on getting this doc change in.

akrabat (approved):
Thanks. It's great to have accurate information in the manual!

PR #927, opened 2020-09-27
Make separations better.

rawsonj:
Add style support for separations.
Add color for separations.

PR #936, opened 2020-10-15
Add dependency check for tests

### LINKED ISSUE #935: Test test_uml_extension fails when plantuml isn't installed
The `test_uml_extension` test fails when the plantuml binary isn't installed. Since it is not a python package it can't be setup as part of the virtual environment as the other optional dependencies can.

Although it would be nice to test the uml functionality of rst2pdf, it seems bad to have the tests fail because the user didn't have plantuml installed, which is only optional.

What would be the best course of action? Should I delete the test? Or could we conditionally disable the test if it is not installed, perhaps issuing a warning?

nfraprado:
Add support for having a `.depends` file alongside a test, with each of its lines listing a dependency. The dependency name is indexed in the `check_dependency` dictionary, returning the method that checks if that dependency is satisfied.
If any of the dependencies listed for a test isn't satisfied, the test is skipped.

This was motivated by the `test_uml_extension`, which requires `plantuml` to be installed. In case this program wasn't present, it would cause the test to fail. Besides adding support for dependencies, this adds the `plantuml` dependency for `test_uml_extension`, so that if `plantuml` is installed it passes, otherwise it is skipped. Fixes #935.

Besides, the first commit simply fixes some Black coding style issues in the file, as the `pre-commit` git plugin wouldn't let me commit otherwise.

### DIFF
--- a/rst2pdf/tests/conftest.py
+++ b/rst2pdf/tests/conftest.py
@@ -41,7 +41,7 @@ def can_run(command_list):
     return _can_run
 
 
-check_dependency = {}
+check_dependency = {'plantuml': can_run(["plantuml", "-pipe"])}
 
 
 def _get_metadata(pdf):
--- a/rst2pdf/tests/conftest.py
+++ b/rst2pdf/tests/conftest.py
@@ -24,6 +24,26 @@ OUTPUT_DIR = os.path.join(ROOT_DIR, 'output')
 REFERENCE_DIR = os.path.join(ROOT_DIR, 'reference')
 
 
+def can_run(command_list):
+    def _can_run():
+        try:
+            p = subprocess.Popen(
+                command_list,
+                stdout=subprocess.PIPE,
+                stdin=subprocess.PIPE,
+                stderr=subprocess.PIPE,
+            )
+            p.terminate()
+        except FileNotFoundError:
+            return False
+        return True
+
+    return _can_run
+
+
+check_dependency = {}
+
+
 def _get_metadata(pdf):
     metadata = pdf.metadata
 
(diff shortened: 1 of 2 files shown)


nfraprado:
Not sure why the CI failed... Maybe I have `pre-commit` misconfigured somehow, since running `pre-commit run --all-files` causes changes in 18 .py files...

rst2pdf/tests/conftest.py
```diff
@@ -24,6 +24,23 @@
 REFERENCE_DIR = os.path.join(ROOT_DIR, 'reference')
 
 
+def has_plantuml():
```
akrabat, line 27: Is there a way we can do this without needing a new function for every dependency we want to add?

akrabat:
> Not sure why the CI failed... Maybe I have `pre-commit` misconfigured somehow, since running `pre-commit run --all-files` causes changes in 18 .py files...

This is the same problem as in #927. When we get to the bottom of it there, we'll know what to do here.

rst2pdf/tests/conftest.py
nfraprado: I did that to give flexibility on how to check if a dependency is satisfied, but yes, it would be better to also have a function to handle the common case of just running a command. I have an idea for that using inner functions. I will push it here so we can review it.

nfraprado:
If we're sure we always want to just run a command with parameters to check for a dependency, we could just cut the middle step and have `check_dependency = {'plantuml': ["plantuml", "-pipe"]}`, with that list always being passed to `subprocess.run`. But since I was not sure, I preferred having the dependency string map to an arbitrary function. Then for the common case of running the command we use `check_dependency = {'plantuml': can_run(["plantuml", "-pipe"])}`, with `can_run` returning a function that runs using the passed arguments.

lornajane:
This is really good!! The build failed because of other problems, which are now fixed in #940. Could you rebase this branch to pick up those changes please? Let us know if we should lend a hand with that.

akrabat (approved):
Looks good to me & easily extendable if we find another dependency.

PR #940, opened 2020-10-21
Fix CI tests

akrabat:
The Travis cache has been hiding changes from us. This was first noticed by black, but also PyMuPDF has changed too which breaks the tests.

The tests then started failing in `fuzzy_coord_diff` with a division by zero error. Investigation showed that PyMuPDF now returns more blocks than it did and some of these new blocks have a coordinate of `0` which was then being used in a division and hence causing the division by zero error. Investigating the algorithm for the fuzzy check showed that it was very very tolerant of a change further down the page to the tune of an 8mm difference in location passing the assertion.

The `fuzzy_coord_diff` check has now been tightened up and this has resulted in other tests failing - mostly legitimately from what I can tell, so I have fixed those up too in this branch.

Key changes:

* `fuzzy_coord_diff` will now fail if the coordinates of a given block is more than ~1mm from the reference PDF.
* `fuzzy_string_diff` will not check image metadata. This is a tricky one and required for plantuml. The image meta data return by PyMuPDF is of this style: '<image: DeviceRGB, width 176, height 164, bpc 8>'`. The width and height are the image's dimensions, not the dimensions that the image is rendered at. Given that plantuml creates different sized images between platform and version, the width and height values in this string will always differ and cannot be checked if we want the tests to pass on our contributors' environments. For now, I've just removed the check is both blocks start with `'<image: DeviceRGB`. I'm confident we can iterate on this if we find a reason to.
* The pytest output on failure has more information in it to help determine which block failed and why.
* The Travis lint build is more verbose on failure.
* A number of files have been reformatted by Black.
* A number of tests failed due to the tightened up `fuzzy_coord_diff` check. These have fallen into these broad categories:
  * Pygments has changed some styles from/to bold. e.g. shebang line is now not bold, Python's "print" is now bold.
  * Vertical spacers with no text before them seem to be a different length nowadays. I have no idea why and this may be an actual thing we need to address at some point.
  * Text in SVG images moves around a bit!
  * The size of images created by plantuml differ between platforms and versions of plantuml. I have introduced `:width:` and `:height` to help mitigate this by allowing us to render a consistently sized diagram in our tests.


stephenfin:
Oh, delightful. Will keep an eye on this.

PS: Apologies for not having looked at the black issue yet. I'm aware it's a problem but I'm slammed with the OpenStack Summit and Project Team (Virtual) Gathering happening this week, on top of my normal work/personal duties. If this is still ongoing the week after next, I can definitely help out here.

lornajane (approved):
Definitely an improvement in so many ways! Thanks :)

PR #947, opened 2020-11-06
Rebased version of renamed files

lornajane:
Replaces #929, rebased to take account of other fixes.


lornajane:
Seems to run _almost_ the same number of tests as master does

akrabat (approved):

PR #950, opened 2020-11-20
Bump travis to Focal Fossa 20.04

### LINKED ISSUE #941: Review Travis version of Ubuntu
**Description of problem**

Travis uses xenial as the ubuntu distro, which is 16.04. Look into using either 18.04 or 20.04 for the test platform.

lornajane:
Let's try to upgrade our travis platform to something users might actually be using. Fixes #941

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 ---
 filter_secrets: false
 os: linux
-dist: xenial
+dist: focal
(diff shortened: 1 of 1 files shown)


akrabat:
I think we should just move to GH Actions at this point. PR #951 does this, so this can be closed.

PR #951, opened 2020-11-27
Try GitHub actions as a CI platform

lornajane:
Since we need to migrate from travis-ci.org to travis-ci.com anyway, how about migrating to GitHub actions instead? This PR attempts to do that (looks like you can see the results on my fork).


akrabat:
@lornajane PR [#2](https://github.com/lornajane/rst2pdf/pull/2) is waiting for you. I worked out which test was missing - `plantuml` needed to be installed. Also limited lint to one Python and enabled running the action when a pull request is raised.

akrabat (approved):
This is great. Thanks!

PR #952, opened 2020-11-27
tests/rename_for_verbosity_pt1 - rename issue test names for verbosity

BinaryKitten:
Renamed the tests for issues < 100 so that they're easier to understand in logs when they fail.

Also updated test_issue_68_1 & test_issue_68_2 so that they refer to the renamed files.


BinaryKitten:
realised i renamed the output files instead of reference. 
I'll rename the reference files and commit.

I am getting different output on 68_1 from reference file - not sure if intentional but going to revert the content change on those two so that there's parity.

akrabat:
Thanks for this work. With the merging of #951, you'll need to rebase against `master` in order to pass the CI checks.

lornajane (approved):
Such an amazing gift to the project! Thank you!

PR #953, opened 2020-11-27
Update CONTRIBUTING for test filenames

akrabat:
Following RM #947, the test input files now have an `rst` extension. This needs to be reflected in `CONTRIBUTING.rst`. Also, they don't have to be prefixed with `test_`.


stephenfin (approved):

nfraprado:
Oops, I was just about to comment that there's also a .depends (albeit only for the plantuml test) to define dependencies for the test, but as this just got merged, I'll make a separate MR adding that later.

Issue #920, opened 2020-09-24
except code path never reached in createpdf.py:add_extensions


dkriegner:
If I read the code [here](https://github.com/rst2pdf/rst2pdf/blob/228411388f481ea32888b9a958f800f2be2450df/rst2pdf/createpdf.py#L1815) correctly the expcept code is never executed since all ImportError are caught inside the try statement. 

This is somehow unfortunate since the resulting error message in case one attempts to load non-existing extensions is not very relevant 
```
[ERROR] pdfbuilder.py:149 local variable 'module' referenced before assignment
Traceback (most recent call last):
  File "/home/dk/.local/lib/python3.7/site-packages/rst2pdf/pdfbuilder.py", line 107, in write
    createpdf.add_extensions(dummy())
  File "/home/dk/.local/lib/python3.7/site-packages/rst2pdf/createpdf.py", line 1595, in add_extensions
    if hasattr(module, 'install'):
UnboundLocalError: local variable 'module' referenced before assignment
```

I stumbled across this due to the recent removal of pdfrw/vectorpdf. sorry for not providing example code showing the issue, but i hope if you look into your code its clear what I mean.

lornajane:
Thanks for letting us know. I'm not sure we've got enough information here to improve the situation but a test case or even a pull request with a fix would be really welcome.

dkriegner:
Sorry, I am not sure what you mean by not having enough information? 
If you read the code starting [here:](https://github.com/rst2pdf/rst2pdf/blob/228411388f481ea32888b9a958f800f2be2450df/rst2pdf/createpdf.py#L1809)

        _names = [firstname, modname]
        try:
            for _name in _names:
                try:
                    module = import_module(_name, 'rst2pdf')
                except ImportError:
                    continue
        except ImportError as e:
            if str(e).split()[-1].replace("'", '') not in [firstname, modname]:
                raise
            raise SystemExit(
                '\nError: Could not find module %s '
                'in sys.path [\n    %s\n]\nExiting...\n'
                % (modname, ',\n    '.join(sys.path))
            )

You will immediately see that the outer except statement which would cause a clean exit with appropriate error message can never be reached because all `ImportError` are caught by the inner try/except. 

The problem seems was introduced in this [commit](https://github.com/rst2pdf/rst2pdf/commit/c3a7c5c2184f853fcbb61e20829b5c0f9d276c39) by @tbpassin 

Likely it should read (similar to the code structure before the mentioned commit):

        try:
[...]

PR #954, opened 2020-11-28
createpdf: Error out when importing extension fails

stephenfin:
Commit c3a7c5c21 modified how we imported extension modules to add
Python 3 compatibility. Unfortunately this introduced a bug, whereby we
will no longer error out if neither of our two attempts to import the
module, by full path and by module path, succeeds.

Restore this behavior.

Closes: #920

### DIFF
--- a/rst2pdf/createpdf.py
+++ b/rst2pdf/createpdf.py
@@ -791,13 +791,16 @@ class FancyDocTemplate(BaseDocTemplate):
                         flowables.insert(i, f)  # put split flowables back on the list
                 else:
                     if hasattr(f, '_postponed') and f._postponed > 4:
-                        ident = "Flowable %s%s too large on page %d in frame %r%s of template %r" % (
-                            self._fIdent(f, 60, frame),
-                            doctemplate._fSizeString(f),
-                            self.page,
-                            self.frame.id,
-                            self.frame._aSpaceString(),
-                            self.pageTemplate.id,
+                        ident = (
+                            "Flowable %s%s too large on page %d in frame %r%s of template %r"
+                            % (
+                                self._fIdent(f, 60, frame),
+                                doctemplate._fSizeString(f),
+                                self.page,
+                                self.frame.id,
+                                self.frame._aSpaceString(),
+                                self.pageTemplate.id,
+                            )
                         )
                         # leave to keep apart from the raise
                         raise LayoutError(ident)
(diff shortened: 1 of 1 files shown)


rst2pdf/createpdf.py
```diff
@@ -1807,23 +1805,30 @@ def __init__(self):
...
         _names = [firstname, modname]
-        try:
-            for _name in _names:
-                try:
-                    module = import_module(_name, 'rst2pdf')
-                except ImportError:
-                    continue
-        except ImportError as e:
-            if str(e).split()[-1].replace("'", '') not in [firstname, modname]:
+        exc = None
```
akrabat, line 1813: Can we call this variable `import_error` or similar?

rst2pdf/createpdf.py
```diff
@@ -1807,23 +1805,30 @@ def __init__(self):
...
+        for _name in _names:
+            try:
+                module = import_module(_name, 'rst2pdf')
+                break
+            except ImportError as e:
+                exc = e
+        else:
+            if not exc:
```
akrabat, line 1821: Is there any scenario where we have entered this `else` statement _and_ `exc` is `None`? I can't think of one, so am I missing something?

akrabat (changes requested):
Thanks @stephenfin. I had to learn how `for..else` works, but this works much better. Just a minor change and a question.

rst2pdf/createpdf.py
stephenfin: In a general sense, this can be triggered if we iterate through an empty list:

    >>> for x in []:
    ...     pass
    ... else:
    ...     print('foo')
    ...
    foo

That will never happen here so not really, but it still felt "safer"

rst2pdf/createpdf.py
stephenfin: Sure

akrabat:
Looks good. With the merging of #951, you'll need to rebase against `master` in order to pass the CI checks.

lornajane:
I think this is still useful, should I rebase it to get this ready to merge?

stephenfin:
If you have time, sure (in general, feel free to rework/rebase any of my patches when necessary). If not, I'll get back to this next week when I'm back in front of a PC

akrabat (approved):

Issue #906, opened 2020-08-29
Stylesheet code examples are in JSON in the manual, while in RSON in the style files


nfraprado:
Stylesheet code examples are in JSON in the manual, while in RSON in the style files and [examples page](https://rst2pdf.org/examples).
Is there some reason for this?

Even though that is valid since both formats are supported, I think it is confusing, specially since users tend to base their style files on the ones provided from installation and from the examples page, and then use the styling code from the manual to add any additional customization.
So making the formats different in those two places makes them incompatible and confusing to the users.

I think RSON would be the preferred format since it is cleaner, and it is already used in every place other than the manual.

If you agree with my point I can send a PR changing the stylesheet code snippets in the manual to be in RSON instead of JSON.

lornajane:
This is a really good point, thanks for opening this issue! However I know that the style files may get an update (see #876 ) so let's make sure we update the manual style examples after that happens.

I'm pleased you created an issue before doing the work on the pull request - not because it wouldn't have been a definite improvement, but because your work might have been short-lived. If you have opinions on the style format, feel free to chime in on the other issue! We can revisit this after that one has been dealt with.

PR #956, opened 2020-12-03
Move to YAML for stylesheets 

### LINKED ISSUE #876: Replace rson with something more broadly adopted
rst2pdf styles are formatted in a special syntax known as [RSON](https://code.google.com/archive/p/rson/). RSON as a standard was effectively a non-starter and appears to be abandoned. There does not appear to be a good reason to retain this format and I'd like to explore switching to a more standard format. There are a great many markup formats available that would be fit for purpose (and many that are not - JSON and XML are both too verbose and difficult to write for humans). I personally think TOML or YAML would be good options.

To compare, consider this snippet taken from `rst2pdf/styles/murphy.style`:

```
styles:
    pygments-bp: {parent: code, textColor: #007020}
    pygments-c:
        fontName: stdMonoItalic
        parent: code
        textColor: #606060
    pygments-c1:
        fontName: stdMonoItalic
        parent: code
        textColor: #606060
    pygments-cm:
        fontName: stdMonoItalic
        parent: code
        textColor: #606060
```

# YAML

YAML is essentially identical to RSON and would make for an easier transition. The `pyyaml` library looks like a good option. However, the YAML spec is notoriously complex and it can be difficult to parse. This likely isn't an issue here but it's worth noting.

```yaml
---
styles:
  pygments-bp: {parent: code, textColor: #007020}
  pygments-c:
    fontName: stdMonoItalic
    parent: code
    textColor: "#606060"
  pygments-c1:
    fontName: stdMonoItalic
    parent: code
    textColor: "#606060"
```

# TOML

TOML is much simpler than YAML and would allow us to integrate with the `pyproject.toml` file introduced in [PEP-518](https://www.python.org/dev/peps/pep-0518/). However, it's quite different from RSON and would require more effort on the users behalf to migrate styles.

```toml
[styles]

  [styles.pygments-bp]
  parent = "code"
  textColor = "#007020"

  [styles.pygments-c]
  fontName = "stdMonoItalic"
  parent = "code"
  textColor = "#606060"

  [styles.pygments-c1]
  fontName = "stdMonoItalic"
  parent = "code"
  textColor = "#606060"
```

---

Personally, I would prefer TOML if we were starting from scratch or if we provided solid documentation (or even a tool) to migrate styles. However, YAML might be the more expedient option. Heck, why not even do both?

lornajane:
Fixes #876 where we've been discussing moving away from RSON to something a bit more widely used/supported/familiar for users.

This PR introduces:
 - Support for stylesheets with `.yaml` and `.yml` extensions, in valid YAML format
 - A converter tool in `rst2pdf/style-to-yaml.py` to convert any existing stylesheet (including our default ones)

Other things to include before merging if we decide this is a useful thing and we should pursue it:
 - stylesheet documentation needs a big overhaul, also see #906 where the manual examples are in JSON at the moment
 - are there other ways to represent some of those nested table styles? They are UGLY
 - what else?

Please kick the tyres, convert your stylesheets, see what you think and let's decide if this is a good approach.


akrabat:
With the merging of #951, you'll need to rebase against `master` in order to pass the CI checks.

lornajane:
Rebased as requested

rst2pdf/styles.py
```diff
@@ -610,7 +611,11 @@ def readStyle(self, ssname):
         fname = self.findStyle(ssname)
         if fname:
             try:
-                return rson_loads(open(fname).read())
+                # Is it an older rson/json stylesheet with .style extension?
+                if fname[-5:] == "style":
```
stephenfin, line 615: Could we use `os.path.splitext` to get the extension instead of this? This was trigger for e.g. `foo.mystyle` too.

rst2pdf/styles.py
```diff
@@ -610,7 +611,11 @@ def readStyle(self, ssname):
...
+                    return rson_loads(open(fname).read())
```
stephenfin, line 616: Could we log a warning here, indicating that users should convert their style files to YAML file?

stephenfin (changes requested):
Some comments inine. Could we get some docs too?

lornajane:
Braindump in case anyone else has batches of styles to convert:

```
for i in *.style; do base=$(basename $i .style); python3 /path/to/style-to-yaml.py $i > $base.yaml; done
```

stephenfin:
> Braindump in case anyone else has batches of styles to convert:
> 
> ```
> for i in *.style; do base=$(basename $i .style); python3 /path/to/style-to-yaml.py $i > $base.yaml; done
> ```

Fwiw, if you renamed the file to use underscores or something else (since dashes aren't valid in a module name), you could do:

```
python -m rst2pdf.style2yaml $i
```

lornajane:
@stephenfin this is useful to know - @akrabat already asked me if we could add the converter as an option to the rst2pdf command itself and this seems relevant for when I get to that point too!

rst2pdf/styles.py
lornajane: I couldn't figure out the "Python" way for this, thanks for the tip!

rst2pdf/styles.py
lornajane: Great idea, added!

lornajane:
A few tests failed, but they pass locally for me (python 3.8). Any idea if we can get the output or figure out how to reproduce the failures so I can take another look?

stephenfin:
Hmm, looks like the python3.7 test failed and the others were cancelled because of this. They likely failed too. Is there any chance you forgot to add some files or something? If not, maybe split this into multiple PRs. One to add the converter, one to add docs, one to convert the test files and a final one to convert the in-tree styles?

lornajane:
This is pretty much working now but I think we should add docs before we merge. I am on the case :)

rst2pdf/style-to-yaml.py
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env python
```
stephenfin, line 1: `python3`

On all Linux distros except for Arch, `python` historically points to `python2.7`. This is changing but it's still safer to be explicit IMO

rst2pdf/style-to-yaml.py
```diff
@@ -0,0 +1,25 @@
...
+'''
```
stephenfin, line 2: style nit: could we use double quotes here? That's more typical for multi-line docstrings.

rst2pdf/style-to-yaml.py
```diff
@@ -0,0 +1,25 @@
...
+    Convert older RSON stylesheets to YAML format
```
stephenfin, line 3: nit: could you dedent this? No reason for it to be indented.

rst2pdf/style-to-yaml.py
stephenfin: Also, license information? Even a SPDX header would do the trick.

```
# SPDX-License-Identifier: MIT
```

rst2pdf/style-to-yaml.py
```diff
@@ -0,0 +1,25 @@
...
+
+import sys
+import json
+import yaml
+from rst2pdf.dumpstyle import fixstyle
+
+from rson import loads as rloads
+
+if __name__ == '__main__':
+    fname = sys.argv[1]
```
stephenfin, line 16: Rather than parsing `argv` directly like this, could we use `argparse`? It's trivial and would allow people to get information on expected input via e.g. `-h`. Also, it would make it easier to convert multiple files at once.

```python3
#!/usr/bin/env python3

"""Convert older RSON stylesheets to YAML format.

Convert the file at PATH from RSON to YAML format.
"""

import argparse
import json
import sys

import yaml

from rst2pdf.dumpstyle import fixstyle

def main():
    parser = argparse.ArgumentParser(
        description=__doc__,
        formatter_class=argparse.RawDescriptionHelpFormatter,
    )
    parser.add_argument(
        'path',
        dest='paths',
        metavar='PATH',
        nargs='+',
        help='An RSON-formatted file to convert.',
    )
    args = parser.parse_args()
    for path in args.paths:
        pass # stuff here...
```

rst2pdf/style-to-yaml.py
```diff
@@ -0,0 +1,25 @@
...
+    # read rson from a file
+    sstr = open(fname, 'rb').read()
```
stephenfin, line 18: You're leaving this file open. You rarely want to use `open` like this. Use a context manager instead:

```python3
with open(fname, 'rb') as fh:
    style_data = fixstyle(rloads(fh.read())
    # ...
```

rst2pdf/style-to-yaml.py
```diff
@@ -0,0 +1,25 @@
...
+    style_data = fixstyle(rloads(sstr))
+    # output the style as json, then parse that
+    json_style = json.dumps(style_data)
+    reparsed_style = json.loads(json_style)
+
+    yaml_style = yaml.dump(reparsed_style)
+    print(yaml_style)
```
stephenfin, line 25: It would be nice to be able to direct this to a file rather than having to do so via the terminal. Perhaps we could have a `--save` flag to dump to `'.'.join((os.path.splitext(fname)[0], 'yaml'))`. If you do this, don't forget to check that the file doesn't exist first and probably error out if so.

rst2pdf/styles.py
```diff
@@ -6,6 +6,7 @@
 import os.path
 import sys
 import re
+import yaml
```
stephenfin, line 9: This is a third-party import and should be grouped with the below.

```
stdlib

third-party

first-party
```

I don't think we've anything enforcing this here but that's what PEP8 suggests iirc.

rst2pdf/styles.py
```diff
@@ -610,7 +611,16 @@ def readStyle(self, ssname):
         fname = self.findStyle(ssname)
         if fname:
             try:
-                return rson_loads(open(fname).read())
+                # Is it an older rson/json stylesheet with .style extension?
+                root_ext = os.path.splitext(fname)
+                if root_ext[1] == ".style":
+                    log.warn(
```
stephenfin, line 617: `log.warn` is deprecated. Use `log.warning`. https://stackoverflow.com/a/15655674

stephenfin (changes requested):
Loads of suggestions for how to make the conversion tool potentially more useful. I do realise this is a single use tool though so you can ignore most of them if you like. The only hard blocker I have is on the use of `log.warn`, which is deprecated and should not be used. If you can change that at least and add the docs you've suggested then I think I'll be happy.

lornajane:
All these comments on the conversion script are helpful! I copied one of our existing files and then made it do what I wanted, I guess I didn't think hard about changing anything that had arrived that worked! A bunch of good advice and I'll have another look

rst2pdf/style-to-yaml.py
lornajane: Yes to the license headers! We must try to unblock your PR for those across the rest of the project too

rst2pdf/style-to-yaml.py
lornajane: Fixed, but half the project uses single ones - decided it made no sense to fix other files in this PR. One for the automated stylechecker tools and a rainy day, perhaps

rst2pdf/style-to-yaml.py
lornajane: Ah ha! Thanks

rst2pdf/style-to-yaml.py
lornajane: oh this is so much nicer now we've turned it more into a CLI thing, and I learned something too!

rst2pdf/styles.py
lornajane: ah, thanks

rst2pdf/styles.py
lornajane: Oops! Good catch.

rst2pdf/style-to-yaml.py
lornajane: I think this is needed especially now I have support for multiple files so easily on the input. It was too much for today but I'll take another look when I can

lornajane: I _think_ I've added that now but feedback always welcome. I mean, it does work! But might not be excellent.

lornajane:
OK I think this is in pretty good shape, can we review again and decide if we can move forward? Main points:
 * Doesn't break anything that exists (that I know about)
 * Conversion utility is ready to use
 * Manual is updated to use new syntax, and has information about migrating stylesheets
 * Tests are also updated
 
There is absolutely more that needs doing on the manual but I tried to mostly move/improve what was there without getting too hung up about the overall state of the manual. As long as I haven't made anything worse then I hope we can proceed with this stylesheets change and keep chipping away at the manual (but I've moved everything so we should probably merge this before anyone does anything else radical with docs)

rst2pdf/style2yaml.py
```diff
@@ -0,0 +1,66 @@
...
+            new_path = '.'.join((os.path.splitext(path)[0], 'yaml'))
+
+            if os.path.exists(new_path):
+                print("File " + new_path + " exists, cannot overwrite")
+            else:
+                print("Creating file " + new_path)
+                with open(new_path, 'w') as file:
+                    file.write(yaml_style)
+        else:
+            print(yaml_style)
```
stephenfin, line 62: Since we can potentially print multiple documents here, can you print a separator before this line so one can identify the different docs? E.g.

```
---
doc1
---
doc2
```

You could also print a comment stating the original filename just after the separator.

More at https://yaml.org/spec/1.2/spec.html#id2760395

rst2pdf/styles.py
```diff
@@ -610,7 +611,16 @@ def readStyle(self, ssname):
         fname = self.findStyle(ssname)
         if fname:
             try:
-                return rson_loads(open(fname).read())
+                # Is it an older rson/json stylesheet with .style extension?
```
stephenfin, line 614: A TODO to remove this when we drop RSON support in the future would be helpful.

stephenfin (approved):
I'm happy with this now. Have two additional nits, but nothing that can't be fixed in a follow-up.

lornajane:
OK, minor points addressed, if either of you are happy @stephenfin @akrabat I think we can merge.

akrabat:
Testing with my presentations, the converted stylesheets generated the correct PDFs, which is nice!

Formatting YAML is clearly hard as none of the options for the `Dumper`'s `default_flow_style` create what I would ideally want which would be lists in `[]` style and dictionaries on separate lines. As that doesn't seem possible, the `default_flow_style` of `None` seems to be the best choice.

stephenfin (approved):

stephenfin:
I guess we should look at releasing a new version now. Does 1.0 make sense yet?

akrabat:
For 1.0, off the top of my head, all I can think of is the error codes one and maybe assessing the way padding/margins work at the end of a page.

Practically, I think that second one isn't something to wait for though.

lornajane:
We said we wanted to update the default stylesheet too, and I'd like us to do that. I'm not sure what else we need to do for a 1.0, but let's start getting our ducks in a row and meeting as many existing PRs as we can include? If the stylesheet update makes it then great. If not, probably not a disaster

PR #959, opened 2020-12-11
Add Python 3.9 to the build matrix and allowing it to fail

lornajane:
This pull request updates the GitHub Actions configuration to add Python 3.9 to the build matrix but allows this new version to fail (which it does, currently). Recommend also updating reportlab for 3.9 compatibility.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8]
+        python-version: [3.6, 3.7, 3.8, 3.9]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


stephenfin (approved):
This looks sane to me

lornajane:
Looking at the build, EVERY test failed on this one! The error is

```
Could not import extension rst2pdf.pdfbuilder (exception: cannot import name 'decodestring' from 'base64' 
```

From the Python 3.9 changelog:

> base64.encodestring() and base64.decodestring(), aliases deprecated since Python 3.1, have been removed: use base64.encodebytes() and base64.decodebytes() instead. (Contributed by Victor Stinner in bpo-39351.)

However I can't figure out where we're using these methods, I think perhaps this could be caused by one of our dependencies. More investigation needed.

lornajane:
Closing this in favour of #969, this PR makes no sense by itself but is being built on nicely over there

PR #960, opened 2020-12-18
Store test outputs to aid debugging

lornajane:
Add an artifact-storing job to the github workflows to store all test outputs. This makes the files available as an "artifact" so we can download them and debug what happened (logs and PDFs). Retains for 8 days.


akrabat (approved):
This is really helpful, thanks!

PR #962, opened 2020-12-23
Fix text formatting

opdavies:
The current markdown text formatting causes the text to show incorrectly
when viewed on GitHub and causes an `Unknown target name` error if you
tried to run rst2pdf on it.

This change fixes the text to use the rst emphasis formatting and fixes
the rendering and compiling issue.


opdavies:
Before:

Selection_001

After:

Selection_002

stephenfin (approved):

PR #965, opened 2020-12-29
Fix crash when generating footnotes out of order under sphinx

nfraprado:
When using rst2pdf with sphinx, if the document has footnotes and the
footnote definition appears before any of the footnote references, it
crashes with the following error:

```
[ERROR] pdfbuilder.py:159 Failed to build doc
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/rst2pdf/pdfbuilder.py", line 156, in write
    docwriter.write(doctree, destination)
  File "/usr/lib/python3.9/site-packages/docutils/writers/__init__.py", line 78, in write
    self.translate()
  File "/usr/lib/python3.9/site-packages/rst2pdf/pdfbuilder.py", line 587, in translate
    self.document.walkabout(visitor)
  File "/usr/lib/python3.9/site-packages/docutils/nodes.py", line 214, in walkabout
    if child.walkabout(visitor):
  File "/usr/lib/python3.9/site-packages/docutils/nodes.py", line 214, in walkabout
    if child.walkabout(visitor):
  File "/usr/lib/python3.9/site-packages/docutils/nodes.py", line 214, in walkabout
    if child.walkabout(visitor):
  [Previous line repeated 2 more times]
  File "/usr/lib/python3.9/site-packages/docutils/nodes.py", line 206, in walkabout
    visitor.dispatch_visit(self)
  File "/usr/lib/python3.9/site-packages/docutils/nodes.py", line 1995, in dispatch_visit
    return method(node)
  File "/usr/lib/python3.9/site-packages/rst2pdf/pdfbuilder.py", line 787, in visit_footnote
    fnr = self.footnotedict[id]
KeyError: 'usage-model_id3'
```

This occurs because the footnote tries to set the text for all
references, even the ones that didn't appear yet. Fix this by splitting
the work, so that the footnote sets the text for the references before
it, and let the later references set their text by themselves.

I still haven't run the tests on this, because I'm having some trouble setting it up... But I'm already posting this so it can be reviewed. I did run pre-commit on this, though.

PR #968, opened 2020-12-30
Add reference PDF for sphinx footnotes test

lornajane:
Missed this from #965 , sorry!


akrabat (approved):
Reference PDF here matches the one built by purest on my local machine 👍

PR #969, opened 2020-12-30
Fix python 3.9 support

nfraprado:
rst2pdf is failing with python 3.9 as already mentioned in #959. Fix it by updating dependencies and adding a fix.

With this there are still a few tests failing:

```

FAILED rst2pdf/tests/input/test_issue_289.rst::test_issue_289
FAILED rst2pdf/tests/input/test_issue_48_thin_plus_thick_table_columns_upset_wordaxe.rst::test_issue_48_thin_plus_thick_table_columns_upset_wordaxe
FAILED rst2pdf/tests/input/sphinx-issue172/conf.py::sphinx-issue172
FAILED rst2pdf/tests/input/sphinx-issue319/conf.py::sphinx-issue319
FAILED rst2pdf/tests/input/sphinx-issue320/conf.py::sphinx-issue320
FAILED rst2pdf/tests/input/sphinx-issue318/conf.py::sphinx-issue318
FAILED rst2pdf/tests/input/sphinx-issue252/conf.py::sphinx-issue252
```

Though the wordaxe one can probably just be deleted as it is no longer supported by rst2pdf.

Since sphinx was updated as part of the requirements.txt update, there were probably big changes so these pdf tests would need to be updated.

At first I tried updating everything but sphinx-related packages, and in that case only the following tests were failing:
```
FAILED rst2pdf/tests/input/test_issue_289.rst::test_issue_289
FAILED rst2pdf/tests/input/test_issue_48_thin_plus_thick_table_columns_upset_wordaxe.rst::test_issue_48_thin_plus_thick_table_columns_upset_wordaxe
FAILED rst2pdf/tests/input/sphinx-issue252/conf.py::sphinx-issue252
```

Since in this case we would still need to fix two tests, I thought we should rather update all 6 above and have a recent sphinx version from now on. So the commit in this PR updates all packages including sphinx.

So before merging this, we need to update the pdf for those 6 tests using sphinx.

Otherwise this works fine in my python 3.9 environment, but let's see it if does as well on older python environments.


nfraprado:
I'll wait for some feedback on whether this is ok on python other than 3.9 and also that updating the dependencies (particularly sphinx) doesn't break everything in older environments, and only then I'll update those failing tests.

lornajane:
This is great! Do you want to merge #959  this branch as well? Then the checks that run on the PR will include Python 3.9 (but that build will be allowed to fail). As things stand, there are no regressions for the older python versions so this is looking very promising!

nfraprado:
Sure, I'll do that. But what I'm worried about is the sphinx version. I think if I update the tests, they will no longer work with older sphinx versions, and I don't know if that's a problem.

nfraprado:
Hm, interesting that even though some tests failed on 3.9, it is shown as succeeded here in Github. Is that intended?

lornajane:
I think we have a problem here. The tests failed but the jobs passed. I _think_ because I tried to force the thing that uploads the output to run regardless of whether the tests failed but it's actually caused the whole thing to look successful when it isn't.

nfraprado:
Okay, so I went back and actually pinned sphinx's version to 1.8.5 and updated the rest. The reason is that (as the version string would indicate) from 1.8.5 to 3.4.1 there were various changes, in particular regarding code domains. I think it would require adding some code to have that correctly working, and only then could we update the sphinx tests that rely on that (the ones showing function definitions, modules, etc. `sphinx-issue318` in particular). So it made more sense to let sphinx be for now and have a separate PR later to have all this sorted out and finally update it.

Regarding the last 3 tests that were failing:

* I ignored `test_issue_289` because although there were differences on the pdf output, the reference pdf formatting was honestly broken, so we should rather ignore it for now and throughly fix it later.
* `test_issue_48_thin_plus_thick_table_columns_upset_wordaxe` was just dropped as wordaxe isn't supported anymore.
* `sphinx-issue252` was failing due to an issue with pillow (https://github.com/python-pillow/Pillow/issues/4225), but as noted there, installing with `pip install --no-binary=Pillow Pillow==8.1.0 --force` made it work.

So with this, python 3.9 is working and all tests are passing, though sphinx continues to use the old 1.8.5 version.

nfraprado:
It seems the CI failure was caused by some weird issue in the infrastructure, rather than the patches. Can someone rerun it?

lornajane:
This is in good shape and I think it's important that we get the 3.9 support as soon as we think it's safe. I'll open separate issues for the 289 test and the sphinx upgrade, they can be looked at separately. The CI system seems confused, the stuff I expected to run did actually run so I'm not sure what else is going on.

We seem to have introduced a conflict so could you rebase please? Then I think we are good to go.

nfraprado:
Rebased. The lxml update to `4.6.2` made by `dependabot` that introduced the conflict was also done by this PR.

One minor detail is that the `requirements.txt` comments annotating where each module comes from that I generated are on separate lines rather than a single one. I'm not sure why that is. I generated it with `pip-compile requirements.in` as instructed in the file. Maybe it's my pip-compile version (5.5.0)? Anyway, just a detail, we can investigate if you think it's worth it.

nfraprado:
Nevermind the above comment, I reformated it all manually with the help of regexes.

akrabat (approved):
This is great work, thanks @nfraprado.

We'll have to do a sphinx upgrade soon as 1.8.5 is far too old!

akrabat (approved):

lornajane:
See also #984 :) Thanks @akrabat!

Issue #958, opened 2020-12-10
rst2pdf crashes with several documents from the Linux Kernel


mchehab:
Trying to build pdf docs from Linux .rst files doesn't work.

**Description of problem**

Cloning this tree, which contains a patch adding rst2pdf support:

   https://git.linuxtv.org/mchehab/experimental.git/log/?h=rst2pdf-v3

and building the documentation with:
```
   $ git clone git://linuxtv.org/mchehab/experimental.git -b rst2pdf-v3
   $ make rst2pdf
```
Should create 71 pdf files under `Documentation/output/pdf` directory. 18 of them have zero size, because rst2pdf crashed.

This is an example of one of such crashes.

```
[ERROR] pdfbuilder.py:159 Failed to build doc
Traceback (most recent call last):
  File "/devel/v4l/docs/sphinx_2.4.4/lib64/python3.9/site-packages/rst2pdf/pdfbuilder.py", line 156, in write
    docwriter.write(doctree, destination)
  File "/devel/v4l/docs/sphinx_2.4.4/lib64/python3.9/site-packages/docutils/writers/__init__.py", line 78, in write
    self.translate()
  File "/devel/v4l/docs/sphinx_2.4.4/lib64/python3.9/site-packages/rst2pdf/pdfbuilder.py", line 672, in translate
    createpdf.RstToPdf(
  File "/devel/v4l/docs/sphinx_2.4.4/lib64/python3.9/site-packages/rst2pdf/createpdf.py", line 659, in createPdf
    pdfdoc.multiBuild(elements)
  File "/devel/v4l/docs/sphinx_2.4.4/lib64/python3.9/site-packages/reportlab/platypus/doctemplate.py", line 1177, in multiBuild
    self.canv.save()
  File "/devel/v4l/docs/sphinx_2.4.4/lib64/python3.9/site-packages/reportlab/pdfgen/canvas.py", line 1277, in save
    self._doc.SaveToFile(self._filename, self)
  File "/devel/v4l/docs/sphinx_2.4.4/lib64/python3.9/site-packages/reportlab/pdfbase/pdfdoc.py", line 219, in SaveToFile
    data = self.GetPDFData(canvas)
  File "/devel/v4l/docs/sphinx_2.4.4/lib64/python3.9/site-packages/reportlab/pdfbase/pdfdoc.py", line 245, in GetPDFData
    return self.format()
[...]

PR #972, opened 2021-01-02
Fix #820: Use right anchor name for target nodes

### LINKED ISSUE #820: Crash on hyperlinks with embedded internal link alias
Reading the rst specification, in particular the section about embedded uris and aliases:

- https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#embedded-uris-and-aliases

I expect the following useful constructs to be supported:
```
Title
=====

This is a `hyperlink with an embedded link alias <Title_>`_.
```

and

```
Oh yes, the _`Norwegian Blue`.  What's, um, what's wrong with it?

This is a `link <Norwegian Blue_>`_.
```

However, they crash rst2pdf with the following traceback:
```
  File "/usr/local/bin/rst2pdf", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python2.7/dist-packages/rst2pdf/createpdf.py", line 1511, in main
    compressed=options.compressed)
  File "/usr/local/lib/python2.7/dist-packages/rst2pdf/createpdf.py", line 570, in createPdf
    elements = self.gen_elements(self.doctree)
  File "/usr/local/lib/python2.7/dist-packages/rst2pdf/basenodehandler.py", line 251, in elemdispatch
    elements = self.getelements(client, node, style)
  File "/usr/local/lib/python2.7/dist-packages/rst2pdf/basenodehandler.py", line 226, in getelements
    elements = self.gather_elements(client, node, style)
  File "/usr/local/lib/python2.7/dist-packages/rst2pdf/basenodehandler.py", line 206, in gather_elements
    return client.gather_elements(node, style=style)
  File "/usr/local/lib/python2.7/dist-packages/rst2pdf/createpdf.py", line 400, in gather_elements
    r.extend(self.gen_elements(n, style=style))
  File "/usr/local/lib/python2.7/dist-packages/rst2pdf/basenodehandler.py", line 251, in elemdispatch
    elements = self.getelements(client, node, style)
  File "/usr/local/lib/python2.7/dist-packages/rst2pdf/basenodehandler.py", line 226, in getelements
    elements = self.gather_elements(client, node, style)
  File "/usr/local/lib/python2.7/dist-packages/rst2pdf/genelements.py", line 188, in gather_elements
    return [Paragraph(client.gen_pdftext(node), style)]
  File "/usr/local/lib/python2.7/dist-packages/rst2pdf/basenodehandler.py", line 281, in textdispatch
    text = self.get_text(client, node, replaceEnt)
  File "/usr/local/lib/python2.7/dist-packages/rst2pdf/basenodehandler.py", line 267, in get_text
    return client.gather_pdftext(node)
  File "/usr/local/lib/python2.7/dist-packages/rst2pdf/createpdf.py", line 390, in gather_pdftext
    for n in node.children])
  File "/usr/local/lib/python2.7/dist-packages/rst2pdf/basenodehandler.py", line 280, in textdispatch
    pre, post = self.get_pre_post(client, node, replaceEnt)
  File "/usr/local/lib/python2.7/dist-packages/rst2pdf/genpdftext.py", line 237, in get_pre_post
    pre = u'<a name="%s"/>' % node['refuri']
  File "/usr/local/lib/python2.7/dist-packages/docutils/nodes.py", line 585, in __getitem__
    return self.attributes[key]
KeyError: 'refuri'
```
They work as expected with, e.g., rst2html.

Expected behavior: no crash, but instead the pdf contains an anchor hyperlink that links the hyperlink to the internal target that the alias references.

**Versions**
Python 2.7.15+
rst2pdf===0.94.1
reportlab==3.5.32
OS: Ubuntu 18.04.3 LTS

nfraprado:
A reference with another reference as its embedded URI yields a target
with either 'refid' or 'refuri' set, depending on whether it points to a
section or an URL, respectively.

Previously it was always assumed to have 'refuri', causing rst2pdf to
crash when the embedded reference was a section. To fix the issue, use
either 'refid' or 'refuri' as the anchor 'name' field, depending on
which is present in the target node.

### DIFF
--- a/rst2pdf/genpdftext.py
+++ b/rst2pdf/genpdftext.py
@@ -257,8 +257,16 @@ class HandleTarget(NodeHandler, docutils.nodes.target):
                 pre = u'<a name="%s"/>' % node['ids'][0]
                 client.targets.append(node['ids'][0])
         else:
-            pre = u'<a name="%s"/>' % node['refuri']
-            client.targets.append(node['refuri'])
+            for attr in ['refid', 'refuri']:
+                value = node.get(attr)
+                if value:
+                    pre = '<a name="%s"/>' % value
+                    client.targets.append(value)
+                    break
+            else:
+                raise ValueError(
+                    "Target node '%s' doesn't have neither 'refid' or 'refuri'" % node
+                )
         return pre, ''
 
 
--- a/rst2pdf/tests/input/test_embedded_reference.rst
+++ b/rst2pdf/tests/input/test_embedded_reference.rst
@@ -1,0 +1,14 @@
+Document title
+##############
+
+See `below <s2idle_resume_>`_, in the `next section <Suspend-to-idle Resume Code
+Flow_>`_.
+
+.. _s2idle_resume:
+
+Suspend-to-idle Resume Code Flow
+================================
+
+See `ioctl() <ioctl_>`_.
+
+.. _ioctl: http://man7.org/linux/man-pages/man2/ioctl.2.html
--- a/rst2pdf/tests/reference/test_embedded_reference.pdf
+++ b/rst2pdf/tests/reference/test_embedded_reference.pdf
@@ -1,0 +1,145 @@
+%PDF-1.4
+%“Œ‹ž ReportLab Generated PDF document http://www.reportlab.com
+1 0 obj
+<<
+/F1 2 0 R /F2 3 0 R
+>>
+endobj
+2 0 obj
+<<
+/BaseFont /Helvetica /Encoding /WinAnsiEncoding /Name /F1 /Subtype /Type1 /Type /Font
+>>
+endobj
+3 0 obj
+<<
+/BaseFont /Helvetica-Bold /Encoding /WinAnsiEncoding /Name /F2 /Subtype /Type1 /Type /Font
+>>
+endobj
+4 0 obj
+<<
+/Border [ 0 0 0 ] /Contents () /Dest [ 7 0 R /XYZ 109.3829 733.0236 0 ] /Rect [ 83.26291 719.0236 109.3829 731.0236 ] /Subtype /Link /Type /Annot
+>>
+endobj
+5 0 obj
+<<
+/Border [ 0 0 0 ] /Contents () /Dest [ 7 0 R /XYZ 195.5429 733.0236 0 ] /Rect [ 142.1829 719.0236 195.5429 731.0236 ] /Subtype /Link /Type /Annot
+>>
+endobj
+6 0 obj
+<<
+/A <<
+/S /URI /Type /Action /URI (http://man7.org/linux/man-pages/man2/ioctl.2.html)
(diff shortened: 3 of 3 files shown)


nfraprado:
A bit more context of why this is an issue, with examples from the kernel documentation (#958).

The following snippet:

```
    `below <s2idle_resume_>`_).

.. _s2idle_resume:

Suspend-to-idle Resume Code Flow
================================
```

should create a link named "below" linking to the "Suspend-to-idle..." section, since the target is right before it. (The same could also be done using the whole section name directly, like ``` `below <Suspend-to-idle Resume Code Flow_>`_ ```)

The generated target node is:

```
<target names="['below']" refid="s2idle-resume"/>
```

which has `refid` pointing to the target, but no `refuri`, which is why the original code crashes here.


Another example, this time that uses `refuri`:

```
of an `ioctl() <ioctl_>`_.

.. _ioctl: http://man7.org/linux/man-pages/man2/ioctl.2.html
```

This creates a link with text `ioctl()` and pointing to the above url.

It generates the following target node:

```
<target names="['ioctl()']" refuri="http://man7.org/linux/man-pages/man2/ioctl.2.html"/>
```

This one indeed has a `refuri`, so it wouldn't crash with the original code. But this example serves to show that just changing the code from `refuri` to `refid` would have solved the first case but would have made this second case crash (which was the first fix I tried).

With the change in this PR, both cases work fine.

rst2pdf/genpdftext.py
```diff
@@ -257,8 +257,13 @@ def get_pre_post(self, client, node, replaceEnt):
                 pre = u'<a name="%s"/>' % node['ids'][0]
                 client.targets.append(node['ids'][0])
         else:
-            pre = u'<a name="%s"/>' % node['refuri']
-            client.targets.append(node['refuri'])
+            for attr in ['refid', 'refuri']:
+                value = node.get(attr)
+                if value:
+                    pre = u'<a name="%s"/>' % value
```
stephenfin, line 263: You don't need the u prefix now that we're Python 3-only.

rst2pdf/genpdftext.py
```diff
@@ -257,8 +257,13 @@ def get_pre_post(self, client, node, replaceEnt):
...
+                    client.targets.append(value)
+                    break
+
```
stephenfin, line 266: Do you want an else block to explicitly error out if neither of these are defined?

stephenfin (changes requested):
This looks sane to me. Can we get a test though? I guess since unit tests aren't a thing yet it'll have to be a reference test, or a modification of an existing reference test. Some small suggestions inline too

rst2pdf/genpdftext.py
nfraprado: Yeah, that would be good. Not sure what type of exception to raise though...

nfraprado: I raised a ValueError. Hope that's a good choice.

stephenfin (approved):
Ta. LGTM now

PR #974, opened 2021-01-03
Document .depends for tests

nfraprado:
#936 added `.depends` for specifying the dependencies of tests, though this was never documented. So, add the documentation for it.

Now that I documented it I realized that it may be too complicated to add a new dependency, since it needs changing the code. Another option would be to specify the whole command line inside the .depends, that way just the `.depends` would suffice.

That said, if we consider that we rarely will add new dependencies for tests, and will more frequently add new tests for already existing dependencies, the current scheme makes it simpler, since just the dependency name is needed in the `.depends`. Thoughts?


stephenfin:
> #936 added `.depends` for specifying the dependencies of tests, though this was never documented. So, add the documentation for it.
> 
> Now that I documented it I realized that it may be too complicated to add a new dependency, since it needs changing the code. Another option would be to specify the whole command line inside the .depends, that way just the `.depends` would suffice.
> 
> That said, if we consider that we rarely will add new dependencies for tests, and will more frequently add new tests for already existing dependencies, the current scheme makes it simpler, since just the dependency name is needed in the `.depends`. Thoughts?

While I'm happy with both approaches, it does seem easier to just keep what we have until we have additional data points (tests) to suggest the effort would be worth it (YAGNI). On that ground, I'm personally happy to merge this as-is, but will hold off to make sure you're okay with keeping this as-is for now.

stephenfin (approved):

akrabat (approved):
I agree with @stephenfin. 

Until we get to the stage where updating `check_dependency` is an unreasonable hassle, it's fine as is.

nfraprado:
Agreed. Let's leave it as it is until we think it's worth the change.
-------- Original Message --------

PR #978, opened 2021-01-09
Add pyyaml to the requirements

nfraprado:
#956 moved stylesheets to yaml, but didn't add `pyyaml` as a requirement. Add it.

This is done on top of #969, since my environment is on python 3.9 and I need it to test the change, and also because when generating the `requirements.txt` with `pip-compile` to add `pyyaml`, it updates other packages much like in #969, so doing it on top of #969 keeps the diff small.

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -9,3 +9,4 @@ sphinx==1.8.5
 svglib
 svg2rlg
 xhtml2pdf
+pyyaml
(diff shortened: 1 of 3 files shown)


lornajane:
Looks good but depends on #969 so blocked until that merges

nfraprado:
Rebased now that #969 merged and since `requirements.txt` changes were conflicting.

lornajane (approved):
good catch, thanks!

PR #979, opened 2021-01-09
Fix embedded references (again)

nfraprado:
#820 didn't actually fix embedded references. It made rst2pdf no longer crash, but with it, using embedded references makes the links point to the wrong place. Using the test edited in this PR as an example:

There is a target on the heading on the second page:
```
.. _s2idle_resume:
```

However, with the "fix" in #820, the following reference:
```
`below <s2idle_resume_>`_
```

that should link to the heading on the second page, produces an anchor named `s2idle_resume` (`<a name="s2idle_resume">`), effectively hiding the heading's anchor and making all other links point to ``` `below <s2idle_resume_>`_ ``` instead, which is on the first page rather than the second.

In this PR, I'm using the `name` property from the node instead of `refuri` so that no longer happens. But I honestly don't even know the purpose of this `get_pre_post()` function, as removing its whole body works just as well...

I also added all different reference types I could think of to the `embedded_reference` test, to make sure everything is working, and also page breaks, so it is actually possible to notice if it doesn't (the original test had everything in a single page, so the issue wasn't visible). Be aware, though, that the test framework doesn't mark it as a fail if the link points to the wrong place.

I'm marking this PR as WIP since, even though it works the way it is, I think we should discuss this over.

### DIFF
--- a/rst2pdf/genpdftext.py
+++ b/rst2pdf/genpdftext.py
@@ -257,16 +257,10 @@ class HandleTarget(NodeHandler, docutils.nodes.target):
                 pre = u'<a name="%s"/>' % node['ids'][0]
                 client.targets.append(node['ids'][0])
         else:
-            for attr in ['refid', 'refuri']:
-                value = node.get(attr)
-                if value:
-                    pre = '<a name="%s"/>' % value
-                    client.targets.append(value)
-                    break
-            else:
-                raise ValueError(
-                    "Target node '%s' doesn't have neither 'refid' or 'refuri'" % node
-                )
+            name = node['names'][0]
+            if name:
+                pre = '<a name="%s"/>' % name
+                client.targets.append(name)
         return pre, ''
 
 
--- a/rst2pdf/tests/input/test_embedded_reference.rst
+++ b/rst2pdf/tests/input/test_embedded_reference.rst
@@ -1,14 +1,42 @@
 Document title
 ##############
 
+
+See `last page`_.
+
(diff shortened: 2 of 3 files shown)


lornajane:
(sorry, pressed the wrong button)

lornajane:
I think we could extend the testing to include links, it looks like PyMuPDF has decent support for that. https://pymupdf.readthedocs.io/en/latest/tutorial.html#inspecting-the-links-annotations-or-form-fields-of-a-page - but I am also happy to have us review this PR as-is, with the extra testing functionality coming in as a separate issue/PR rather than holding things up.

lornajane:
In the absence of further feedback, and a change to master that fixes the failing builds, it's my view that we should:
 * rebase this PR (to pick up the build fixes) and remove the WIP indicator so it can be reviewed as-is
 * open an issue to capture the need for checking links point to the same place, including information about which test this is that already has a bunch of internal links to check on

nfraprado:
What bothers me is that I'm really not sure if this is the right fix, since, as I said, removing the function completely has the same appeared effect of this patch...

That said, this does fix the issue. And perhaps the wrong fix is still better than no fix 😝.

lornajane (approved):
There's a bunch of valid concerns here and as maintainers, we share them. We have the CI set up to help us with regressions, and I think the best thing we can do is what you have done here: fix the things we see and stand by to make further changes as needed. This fix, and especially this test, move the project forward, and I'm really happy to have both the code and the discussion around this!

PR #985, opened 2021-01-10
Remove wordaxe from issue template since we no longer use it

lornajane:
A tiny fix to remove something from our long-winded issue template that we no longer need. Every little helps!


akrabat (approved):

akrabat:
Thanks!

PR #987, opened 2021-01-15
Update CI versions and only upload assets on failure

akrabat:
We require the tests to pass under Python 3.9 now. Python 3.10 is in early dev, but we should test it, allowing it to fail - which it does!

Also only upload the test output files if they fail as we don't need them if the tests pass. It'd be neater to only upload the failed PDF and log files, but this is a good first step.

Finally, split the install dependencies step to separate the apt installation from the pip one as it makes it easier to diagnose when it fails due to fewer lines to scroll through.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,9 +11,6 @@ jobs:
       matrix:
         python-version: [3.6, 3.7, 3.8, 3.9]
         experimental: [false]
-        include:
-          - python-version: 3.10-dev
-            experimental: true
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


lornajane (changes requested):
This looks a lot better in the failed uploads handling! But the 3.10 does not even successfully install the dependencies, and causes the build to look "failed". My worry is that we stop reacting to failed builds because they always fail, and miss something important. I'd prefer us to work on 3.10 separately but adopt the other changes here immediately.

akrabat:
>  I'd prefer us to work on 3.10 separately but adopt the other changes here immediately.

3.10 removed.

nfraprado:
Great! This should hopefully fix the issue I had in https://github.com/rst2pdf/rst2pdf/pull/969#issuecomment-753649200 where the upload in itself failed. This way we never fail due to infrastructure errors when the tests aren't already failing.

lornajane (approved):

lornajane:
Thanks for the quick response :) Definitely need to look at 3.10 sometime but at least we're all green for now.

Issue #884, opened 2020-07-11
Review default style sheet


akrabat:
We should review the default style sheet and maybe improve it.

The current one needs to be available so that people who have based their styles off of it can just add `-s {old name}.style` in order to continue getting what they expect. (We'll obviously need com up with a name for the old default style - preferably one that lends itself to becoming a series.

We should keep the current default styles as `legacy` so that people can build their current docs.

lornajane:
The current default stylesheet produces `document1.pdf` attached and I have also included `document`1-pandoc.pdf` to show the default pandoc output (pretty sure this is not what we are aiming for though!).

(Created from the `document1.rst` source file in the website project)
[document1.pdf](https://github.com/rst2pdf/rst2pdf/files/5792661/document1.pdf)
[document1-pandoc.pdf](https://github.com/rst2pdf/rst2pdf/files/5792663/document1-pandoc.pdf)

nfraprado:
Nice comparison.

I can see that rst2pdf's page is slightly skinnier than pandoc's. Pandoc's looks more like an A4 which should be the default for rst2pdf IMO.

Another thing is that the document title should probably have more space after it, like pandoc.

The big margins and serif fonts in pandoc's document gives it a LaTeX/Scientific Paper feel, which I wouldn't think would be a good default for rst2pdf, though it would be nice to include that as a built-in style.

lornajane:
I started looking at some options, the default is A4 but the margins make it seem completely different! There are also really really limited choices on fonts, so how about the attached?
[document1.pdf](https://github.com/rst2pdf/rst2pdf/files/5845601/document1.pdf)

nfraprado:
That's a great start, although I'm not fond of the gray colored titles, as they seem too washed out.

I think we should really test this using a more complete sample document, with multiple pages, sections, and at least the most used rst constructs. There's only so much we can check in a simple document like that one.

habamax:
You could also check asciidoctor-pdf defaults for inspiration.

## Without title page 

```
= PDF Documents: From your Text Editor

First para is Abstract and has a bigger font by default.

Lorem ipsum dolor sit amet, consectetur adipiscing elit.
Maecenas feugiat fermentum pretium. Cras eu dolor imperdiet justo mattis
pulvinar. Cras nec lectus ligula. Proin elementum luctus elit, a tincidunt quam
facilisis non. Nunc quis mauris non turpis finibus luctus. Maecenas ante
[...]

PR #988, opened 2021-01-20
Stylesheet Refresh

lornajane:
Marked as draft because it's intended as a starting point for #884  and changing the default stylesheet will probably cause a lot of test failures (we might want to either update the tests or have the existing ones use the old stylesheet for now) and there are probably lots of things I didn't think about yet. Feel free to fork and counter-PR with your own ideas; I am also open to adding/culling styles while we're at it.


akrabat:
I like the use of a the different font for the headings. Much nicer. I also like the removal of bold on the table of contents. 

As you've renamed all the font aliases, I'm not sure that overrides to the old names would be picked up any more. Therefore, I think we should provide the current default stylesheet as `classic` or something so that users can do `-s classic` on all their existing projects.

lornajane:
I'll push the old stylesheet (I thought I had but diff says no!) definitely and we will need to decide what to do about updating the tests or setting them to use the old stylesheet. With the fonts, now I've seen "behind the curtain" I am not sure I can move on and leave it as it is so there are probably more changes to come on this before the PR is ready.  I _think_ with some changes we can get the default fonts to work out-of-the-box even when packaged as a snap, and allow embedding to workaround the other errors.

akrabat:
Some of the font names had the previous `std` prefix. Updated to use to the `font` prefix in https://github.com/rst2pdf/rst2pdf/pull/988/commits/9e4429864179ad7fe1e3757540c0dbbf8ce6bae7

akrabat:
I have a few colour change suggestions. Mostly because I don't like beige!

* A light grey background for code blocks
* White on dark gray for table headers
* A warmer yellow background for admonitions and sidebar

See https://github.com/lornajane/rst2pdf/pull/4/files for the diff

Thoughts?

This is a test PDF: [testing.pdf](https://github.com/rst2pdf/rst2pdf/files/6056814/testing.pdf)

Created using this input: [testing.txt](https://github.com/rst2pdf/rst2pdf/files/6056815/testing.txt)

nfraprado:
I really liked the gray on code and tables a lot better! (I also dislike beige though, so I might biased :P)

On a side-note, from your example, in the code block, the `def` seems to be using a different font (or at least a different size) from `myFun`. Not sure if it's intended, or if it's an unrelated issue, but just wanted to point it out.

akrabat:
> On a side-note, from your example, in the code block, the `def` seems to be using a different font (or at least a different size) from `myFun`. Not sure if it's intended, or if it's an unrelated issue, but just wanted to point it out.

So it is! I wonder why that is…

akrabat:
> On a side-note, from your example, in the code block, the `def` seems to be using a different font (or at least a different size) from `myFun`. Not sure if it's intended, or if it's an unrelated issue, but just wanted to point it out.

This is because the pygments styles define fonts using the `std` aliases which haven't been set in the new `styles.yaml`. 

The easiest thing to do is to use the old font aliases names if they are less obvious.

nfraprado:
I think carrying on with the change from `std` to `font` and updating `pygments2style.py` makes the most sense. These font names are used by the users, so should be as intuitive as possible. And that change doesn't seem like too much extra work.

lornajane:
This doesn't look like much but should update existing styles to use the new font naming, and improves the font finding/loading a little. It's been quite an expedition. The tests, clearly, basically all fail at this point. I'm working through with a hack to put the old stylesheet into every test to try to work out how many of these are broken by these code changes, not just by style changes (first impressions: the sphinx tests look really unhappy) and also to fix anything I missed in creating the `rst2pdf-0-9` stylesheet. After that, I'd prefer us to update the reference PDFs to use the new stylesheet, I realise that's an undertaking!

akrabat:
Why does it say that all checks are passing even though the tests have failed when you look at the details?

lornajane:
Hmm. I think it's because we allow the tests to fail and still progress to the next step which is archiving the output. Although I must admit I thought we'd talked about it when we were working on that feature ...

lornajane:
OK I think I'm going to admit that I'm stuck on improving the fonts situation. What we do have:
 - better font finding/loading
 - intentionally separate fonts for headings rather than using some sort of bold face hack (it still makes a sans/serif assumption but at least it doesn't overload the same font names)
 - styles and pygments style generator uploaded
 
What we don't have:
 - the ability to always and reliably work out what formatting should be used for bold fonts under every possible font loading scenario. We can't know whether we should update `fontSansBold` in every situation, especially where people might have migrated different stylesheets. People need to define a whole set of fonts ideally if they want to use custom fonts; the out-of-the-box should be fine at this point though.
 - migration/upgrade instructions. I've split this into #1010 

Next steps:
 - if there's something else we need to do before we can review and release this, identify it
 - regenerate reference PDFs to use the new styles. My theory is that if I can run the test on this branch but with the old stylesheet and it passes, then I can run the test with the new stylesheet and use that as our reference. Anything that fails beyond that test needs more attention.

I expect this to be the last thing required before a 1.0 release, so I'm keen to keep it moving.

lornajane:
Here's how the manual looks now with the current stylesheet (if you have other test documents, go ahead and try them, see what you think or share them!)
[manual.pdf](https://github.com/rst2pdf/rst2pdf/files/6291459/manual.pdf)

akrabat:
> regenerate reference PDFs to use the new styles. My theory is that if I can run the test on this branch but with the old stylesheet and it passes, then I can run the test with the new stylesheet and use that as our reference. Anything that fails beyond that test needs more attention.

Makes sense to me.

lornajane:
Pull requests that change 300 files (and cause the build to not even run) are the best! `</sarcasm>` Definitely a work in progress but I have less than 20 locally failing tests now, and I might need some help on the last mile.

akrabat:
Merged master into this branch to try and get GH actions to run.

akrabat:
Hopefully that set of updates passes on more computers than mine!

I've done a mix of updating the reference PDF after inspection and updating the tests's style file to make it match the current reference PDF. 

I've also added a couple of explicit error messages for when a template is missing or if the template is has no frames which makes it easier to work out why rst2pdf failed to create a PDF in those cases.

lornajane:
I'm not sure how we're going to review this thing but I think it's done, and the tests all pass, so .... can anyone take it for a spin, please?

akrabat (approved):
I'm good with this. We can iterate as and when needed.

Issue #550, opened 2015-08-26
Traceback on malformed rst


valholl:
As reported on https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=639544 when using rst2pdf on malformed rst that causes a docutils.utils.SystemMessage exception, rst2pdf dies with a traceback instead of a nicer error message.

I've checked that this is still happening in master.

```
$ ./env/bin/rst2pdf fail.rst 
fail.rst:15: (SEVERE/4) Title level inconsistent:

::
..
Traceback (most recent call last):
  File "./env/bin/rst2pdf", line 9, in <module>
    load_entry_point('rst2pdf==0.93.dev-r0', 'console_scripts', 'rst2pdf')()
  File "/home/valhalla/tmp/rst2pdf/env/local/lib/python2.7/site-packages/rst2pdf-0.93.dev_r0-py2.7.egg/rst2pdf/createpdf.py", line 1495, in main
    compressed=options.compressed)
  File "/home/valhalla/tmp/rst2pdf/env/local/lib/python2.7/site-packages/rst2pdf-0.93.dev_r0-py2.7.egg/rst2pdf/createpdf.py", line 540, in createPdf
    settings_overrides=settings_overrides)
  File "/usr/lib/python2.7/dist-packages/docutils/core.py", line 479, in publish_doctree
    output = pub.publish(enable_exit_status=enable_exit_status)
  File "/usr/lib/python2.7/dist-packages/docutils/core.py", line 217, in publish
    self.settings)
  File "/usr/lib/python2.7/dist-packages/docutils/readers/__init__.py", line 72, in read
    self.parse()
  File "/usr/lib/python2.7/dist-packages/docutils/readers/__init__.py", line 78, in parse
    self.parser.parse(self.input, document)
  File "/usr/lib/python2.7/dist-packages/docutils/parsers/rst/__init__.py", line 172, in parse
    self.statemachine.run(inputlines, document, inliner=self.inliner)
  File "/usr/lib/python2.7/dist-packages/docutils/parsers/rst/states.py", line 171, in run
    input_source=document['source'])
  File "/usr/lib/python2.7/dist-packages/docutils/statemachine.py", line 239, in run
    context, state, transitions)
  File "/usr/lib/python2.7/dist-packages/docutils/statemachine.py", line 460, in check_line
    return method(match, context, next_state)
  File "/usr/lib/python2.7/dist-packages/docutils/parsers/rst/states.py", line 2962, in text
[...]

PR #990, opened 2021-01-23
Add try/catch for errors such as incorrect title ordering

lornajane:
From #550, catch the exceptions that bubble up so we can explain and exit rather than throwing the traceback. Too much output makes the actual error easy to miss so I hope this helps a little.

Things I would like to add but I'm not sure how:
 * A test that tests the error output so I can put the test from #550 in and check that it keeps doing this
 * The output has strange dots and colons in it, output between consecutive lines in createpdf.... I don't think we should block on this but I would love to know how to tidy it up!


akrabat:
I think that the best way to address the testing of this is to ensure that this situation results in a non-zero return code which we'll be able to test for when #886 is in place.

akrabat:
The strange dots and colons comes from docutils. This is the Exception when inspected in the debugger:

`'...rst2pdf/tests/input/test_malformed_rst.rst:15: (SEVERE/4) Title level inconsistent:\n\n::\n..'`

We get it twice in the log as docutils logs it and then we do.

akrabat:
I've rebased to pick up #925 now that it's merged to develop. Also updated to set the return code on failure(https://github.com/rst2pdf/rst2pdf/pull/990/commits/e816ad985218d0b919faf074a5dbb338d821beaa), add a test(https://github.com/rst2pdf/rst2pdf/pull/990/commits/3406b7ccdee5c1e61d40bb97c1c48a43d74616dc) and removed the duplicate log entry(https://github.com/rst2pdf/rst2pdf/pull/990/commits/8b460a4fe10ea578f6f99f79638415dd12890670)

The output is much neater now:

```
$ rst2pdf input/test_malformed_rst.rst -o output/test_malformed_rst.pdf
input/test_malformed_rst.rst:15: (SEVERE/4) Title level inconsistent:

::
..
[ERROR] createpdf.py:547 Cannot generate PDF, exiting
```

@lornajane If you could review my changes and don't find anything, then I think this is good to merge.

lornajane:
@akrabat looks great but I can't review "my" pull request officially! Want to review and merge?

akrabat (approved):
Lorna's changes look good to me and she's happy with my updates.

PR #992, opened 2021-02-13
Make sure table columns are wide enough to use

### LINKED ISSUE #967: Small table overflowing page
A small table is considered too large for any frame.

**Description of problem**

Building the pdf for the following document:

```
PDF Documents From Your Text Editor
###################################

=  =======================================
0  Register as a non-blocking blk-mq drive
1  Register as a blocking blk-mq driver de
=  =======================================
```

Gives the following error:

```
  File "/usr/bin/rst2pdf", line 8, in <module>
    sys.exit(main())
  File "/usr/lib/python3.9/site-packages/rst2pdf/createpdf.py", line 1683, in main
    RstToPdf(
  File "/usr/lib/python3.9/site-packages/rst2pdf/createpdf.py", line 659, in createPdf
    pdfdoc.multiBuild(elements)
  File "/usr/lib/python3.9/site-packages/reportlab/platypus/doctemplate.py", line 1167, in multiBuild
    self.build(tempStory, **buildKwds)
  File "/usr/lib/python3.9/site-packages/reportlab/platypus/doctemplate.py", line 1080, in build
    self.handle_flowable(flowables)
  File "/usr/lib/python3.9/site-packages/rst2pdf/createpdf.py", line 813, in handle_flowable
    self.handle_frameEnd()
  File "/usr/lib/python3.9/site-packages/reportlab/platypus/doctemplate.py", line 726, in handle_frameEnd
    self.handle_pageEnd()
  File "/usr/lib/python3.9/site-packages/reportlab/platypus/doctemplate.py", line 668, in handle_pageEnd
    raise LayoutError(ident)
reportlab.platypus.doctemplate.LayoutError: More than 10 pages generated without content - halting layout.  Likely that a flowable is too large for any frame.
```

Even though the table is small, both in width and in height, it is considered too large for any page and causes rst2pdf to abort.

Removing a single character from the table, it now becomes small enough to fit the page and the pdf is generated. But seeing the result pdf, it is clear that the table should have fitted with that extra character:

Table working

🖥 **Versions**

`python -V`: 3.9.1

`pip freeze | grep rst2pdf`: 0.98

`pip freeze | grep reportlab`: 3.5.57

Which operating system are you using? Archlinux

lornajane:
Fixes #967  where a table overflows the page. The problem isn't that the table is too large, it's that the column width is smaller than its padding and reportlab doesn't know how to calculate if it will fit in a space when the number is negative so it returns huge dimensions and it overflows.

This PR just makes any column less than 4% of the table be 4%. Sometimes, our tables will be more than 100% as a result, but that hasn't seemed to upset any of the tools. Weirdly none of the tests fail with this change. I'm not sure if we should "bodge" in this way, if it's a useful way to get probably-working tables for most people out of the box, or just one special hack too many. The PR is here for discussion.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install apt dependencies
       run: |
-        sudo apt-get -y install plantuml libxml2-dev libxslt-dev python3-dev
+        sudo apt-get update && apt-get -y install plantuml libxml2-dev libxslt-dev python3-dev
     - name: Install pip dependencies
       run: |
         python -m pip install --upgrade pip
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install apt dependencies
       run: |
-        sudo apt-get update && apt-get -y install plantuml libxml2-dev libxslt-dev python3-dev
+        sudo apt-get update && sudo apt-get -y install plantuml libxml2-dev libxslt-dev python3-dev
     - name: Install pip dependencies
       run: |
         python -m pip install --upgrade pip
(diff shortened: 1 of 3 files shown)


lornajane:
Build fails because apt can't install Java, I'm going to hope that fixes itself and kick off the build again tomorrow

lornajane:
The extra commits here are to update apt before installing dependencies, since it was causing build failures. This should go in separately unless we're merging this PR imminently

akrabat (approved):
Great sleuthing!

PR #994, opened 2021-03-01
Fix writing to stdout

### LINKED ISSUE #991: rst2pdf fails to write output to stdout
Hello, the following command gives me an error:

```
rst2pdf test.rst -o -

  File "/home/olivier/.virtualenvs/tucc/bin/rst2pdf", line 8, in <module>
    sys.exit(main())
  File "/home/olivier/.virtualenvs/tucc/lib/python3.8/site-packages/rst2pdf/createpdf.py", line 1683, in main
    RstToPdf(
  File "/home/olivier/.virtualenvs/tucc/lib/python3.8/site-packages/rst2pdf/createpdf.py", line 659, in createPdf
    pdfdoc.multiBuild(elements)
  File "/home/olivier/.virtualenvs/tucc/lib/python3.8/site-packages/reportlab/platypus/doctemplate.py", line 1177, in multiBuild
    self.canv.save()
  File "/home/olivier/.virtualenvs/tucc/lib/python3.8/site-packages/reportlab/pdfgen/canvas.py", line 1277, in save
    self._doc.SaveToFile(self._filename, self)
  File "/home/olivier/.virtualenvs/tucc/lib/python3.8/site-packages/reportlab/pdfbase/pdfdoc.py", line 222, in SaveToFile
    f.write(data)
TypeError: write() argument must be str, not bytes
```

I obtained the same error using this command:

```
cat test.rst | rst2pdf 
```

The content of my test.rst file is:

```
Title
###########

Header
============

some text 
```

```
python3 --version
Python 3.8.5
rst2pdf --version
0.98
```

thank you
Best regards
Olivier

akrabat:
After enquiring on the [ReportLab mailing list](https://pairlist2.pair.net/pipermail/reportlab-users/2021-February/012042.html), the correct way to pass a valid stdout stream to ReportLab is to use `sys.stdout.buffer`. I updated the code to send this and it works.

Testing it is a little hard. For now, I simply check that we don't crash (return a non-zero status code) when running, but really we could do with checking the data in the test's log file is what we expect as this is the PDF file created.

Fixes #991.


stephenfin (approved):
Played around with this. This change looks reasonable. Still not sure about the test but I can't think of anything better. Did consider doing this via normal unit tests outside of the custom test harness we have, but tbh this is still an integration-style test so what you've done makes as much sense.

lornajane (commented):
This looks great! Merged with a tiny note to our future selves in the source of the test :)

akrabat:
> This looks great! Merged with a tiny note to our future selves in the source of the test :)

Thanks! That's very wise 👍

PR #998, opened 2021-03-20
Update dependencies

lornajane:
I'm not sure if the automated dependency updates are new but this PR combines all three #995 #996 #997 that came in over the last couple of days, basically this seemed like a simpler way to test all the changes rather than rebasing each of the PRs in turn before merging.

The automated tool does entirely reformat the `requirements.txt` file, I can't think why that could cause a problem but I'm mentioning it just in case!

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,48 +4,115 @@
 #
 #    pip-compile requirements.in
 #
-aafigure==0.6             # via -r requirements.in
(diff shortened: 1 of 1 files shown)


akrabat (approved):
That reformatting is weird isn't it? I guess it'll go back to the other way when we next use pip-compile!

PR #1000, opened 2021-03-29
Update lxml to 4.6.3

akrabat:
Security update as noticed by dependabot, but no PR seemed to be created.


akrabat:
I was too fast! Dependabot got there in #1001

PR #1006, opened 2021-04-06
Add a (temp) fix for a change in black branch naming

lornajane:
The build errors we saw in the last few days were due to the Black project removing their `stable` branch which is the recommended one to use (see https://github.com/psf/black/issues/2079). This change should fix the build, but we will need to either bump the version number or return to a tracking branch at some future time.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,7 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


stephenfin (approved):
Looks good to me

akrabat:
I think `stable` is back as per: https://github.com/psf/black/issues/2079#issuecomment-814567630

lornajane:
ha! I tried holding out for a bit and then decided we couldn't wait :)

PR #1017, opened 2021-08-03
Fix: make pdf not working on Sphinx 4

gXiah:
Fix: make pdf not working on Sphinx 4.

Few changes to make the ``make pdf`` command work with Sphinx 4

### DIFF
--- a/rst2pdf/pdfbuilder.py
+++ b/rst2pdf/pdfbuilder.py
@@ -228,7 +228,10 @@ class PDFBuilder(Builder):
             # ALL the documents data, not just this one.
             # So, we preserve a copy, use just what we need, then
             # restore it.
-            t = copy(self.env.indexentries)
+            
+            tmp = self.env.domains["index"]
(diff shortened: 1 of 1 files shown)


lornajane:
Could you say a bit more about what this fixes and how to test it please? It looks like this fixes something from sphinx but I'm not sure how to review it from an rst2pdf perspective.

lornajane:
The build is failing on the pre-commit hook, the contributing file gives information about how to run these checks https://github.com/rst2pdf/rst2pdf/blob/master/CONTRIBUTING.rst#run-checks-locally

akrabat:
Sphinx 4 does not have an `env.indexentries` and instead has `env.domains["index"].entries`, so this has to be accounted for before `IndexEntries()` is called.

Similarly, `doc2path()` works differently now and so needs different code.

I'm closing this as it doesn't work as is. Feel free to update and reopen though.

PR #1023, opened 2021-08-30
Add tests for Sphinx pdf_break_level config option

akrabat:
This is a set of tests that show how the `pdf_break_level` option affects the same document.

 As useful for documentation as for proving it works, I think but equally it could be argued that this would be better in the manual or on the website.


lornajane (approved):
clear and useful, thanks!

lornajane:
I know what you mean about the documentation but both we and our users are more likely to see any regression or problem if it's in the test suite ...

PR #1041, opened 2021-12-07
Exit with status code 1 on font loading failure

### LINKED ISSUE #1034: Fonts referenced with directory name are not used and there is no error
For the latest master (4ecc41c6379ebe8668d795b240640c235983a278), if you have fonts in a subdirectory and reference the font with the subdirectory name, then it will not be used and no error message will be displayed.

i.e. consider this stylesheet:

```
embeddedFonts:
  - [fonts/GenBasR.ttf, fonts/GenBasB.ttf, fonts/GenBasI.ttf, fonts/GenBasBI.ttf]
fontsAlias: {stdFont: fonts/GenBasR, stdBold: fonts/GenBasB}
```

rst2pdf will not use `GenBasR` for the `stdFont`, and will not display an error message.

**What is the expected output?**

An error about not finding the font should be displayed.

akrabat:
This PR deals with two situations related to font matching:

Firstly, if the specified embedded font cannot be loaded, this is now treated as an error and rst2pdf will exit with a status code of 1.

Secondly, rst2pdf will now output a warning if fc-match substitutes a different font to the one that was asked for. This situation happens if the font name specified hasn't been embedded as fc-match will then look for it. As fc-match will always return something, the user may get a completely incorrect font and not realise what's happened. Hence we now display a warning when this happens.

This is a warning, not an error as the font selected by fc-match may very well be good
enough for the user and so the PDF should be created.

Fixes #1034


akrabat:
CI issues fixed by running `pre-commit clean` locally… Turns out that I had a stale cache.

akrabat:
More work to do on this one now that I've understood the problem better…

PR #1048, opened 2021-12-15
Support multiple style names for the `class` directive

### LINKED ISSUE #1035: Multiple classes on a node isn't handled
**Description of problem**

If a node has multiple classes assigned to it, then only the first class will actually be respected. For very manually generated files this is probably fine, but it quickly breaks down with a more complex system such as Sphinx.

Sphinx will generate a whole bunch of classes on many elements, and it is rarely the most specific that is the first assigned class. One example is the whole domain system in Sphinx where relevant nodes get the two classes `<class>` and `<class>-<type>`. With rst2pdf you then only get to style `<class>`, which means that you lose the ability to style different types.

I'm not sure how things should ideally work. Something as complex as CSS' selector and inheritance system is probably overkill. But something needs to be done to allow handling of multiple classes.

**What is the expected output? What do you see instead?**

Right now nodes get styled after their first specified class. I'd like more control over which class is used to map to a style when there are multiple.

🖥 **Versions**

 * Python 3.6.12
 * rst2pdf 0.98
 * reportlab 3.6.1
 * Sphinx 4.2.0

akrabat:
For each style after the first one listed, apply the attributes that are different from the defaults to the compbined style. This results in a style that's the combination of all the listed styles.

This has been  long-standing wish list item for myself!

Addresses #1035

### DIFF
--- a/rst2pdf/tests/input/test_multiple_styles.rst
+++ b/rst2pdf/tests/input/test_multiple_styles.rst
@@ -1,0 +1,41 @@
+Multiple styles test
+====================
+
+This test ensures that multiple classes can be applied to an element and that order matters.
+
+Class definitions:
+
+.. code-block:: yaml
+   :include: test_multiple_styles.yaml
+
+|
+
+``.. class:: foo``:
+
+.. class:: foo
+
+This is red, left aligned and large.
+
+|
+
+``.. class:: bar``:
+
+.. class:: bar
+
+This is blue, centre aligned and small.
+
+|
+
+``.. class:: foo bar``:
+
+.. class:: foo bar
+
+This is blue, centre aligned and large.
+
+|
+
+``.. class:: bar foo``:
+
+.. class:: bar foo
+
+This is red, centre aligned (as ``foo`` doesn't specify aligment)  and large.
--- a/doc/manual.rst
+++ b/doc/manual.rst
@@ -249,6 +249,15 @@ You can style paragraphs with a style using the class directive::
 
   This one is not.
 
+Multiple styles can be listed and are applied in order::
+
+  .. class:: special redtext
+
+  This paragraph is special and is red.
+
+  This one is not.
+
+
 Or inline styles using custom interpreted roles::
 
    .. role:: redtext
--- a/doc/manual.rst
+++ b/doc/manual.rst
@@ -249,11 +249,11 @@ You can style paragraphs with a style using the class directive::
 
   This one is not.
 
-Multiple styles can be listed and are applied in order::
+Multiple styles can be listed and are applied in order where properties in the right hand styles override those to the left::
 
-  .. class:: special redtext
+  .. class:: special bluetext redtext
 
-  This paragraph is special and is red.
+      This paragraph is special and is red.
 
   This one is not.
 
--- a/rst2pdf/basenodehandler.py
+++ b/rst2pdf/basenodehandler.py
@@ -31,6 +31,7 @@ type of docutils node, then default processing will occur and a warning
 will be logged.
 '''
 
+from copy import copy
 import inspect
 import types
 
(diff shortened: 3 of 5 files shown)


CendioOssman:
I haven't tried actually running this change, but I see two questions directly:

 * Should you perhaps reverse the order? Without this PR the first class determines everything. I.e. it has absolute priority. With the suggestion in this PR any conflicts will be resolved in favour of the later classes. i.e. the last class has priority, not the first. Reversing the order would then seem like the more backward compatible approach.
 * How will this hold up during inheritance? Are things considered non-default if a base class overrides things? E.g. both `Foo` and `Bar` have `Base` as their parent. `Foo` has `textColor: red`, `Bar` doesn't have `textColor` specified and `Base` has `textColor: green`. What color will I get if I specify `.. class:: Foo Bar`? Will the inherited green color trump the red? And what if the inheritance is not from a common base?

I'm afraid I don't have any immediate good answers for these, but it's probably best to have these well defined before merging this as they will be difficult to change once people are relying on them.

akrabat:
> * Should you perhaps reverse the order? Without this PR the first class determines everything. I.e. it has absolute priority. With the suggestion in this PR any conflicts will be resolved in favour of the later classes. i.e. the last class has priority, not the first. Reversing the order would then seem like the more backward compatible approach.

This is a really good question. I'm not more worried about BC vs intuitive and logical going forward. i.e. I don't want to commit to something that the majority of users find confusing and is hence an ongoing problem going forward just to keep BC on something we've never supported.

In terms of "intuitive and logical", my thought was that  "stacking" from left to right felt most obvious to my eyes.

e.g. when reading this RST:

```rst
.. class:: redtext bluetext

Some text
```

I personally would guess that the "Some text" string would be blue (assuming class names represent what they do).


When considering multiple style classes applied text, the most obvious analogy that users will be familiar with is HTML, where the order of definition determines the "winner" rather than order of application on the element.

i.e. for this HTML:

```html
<html>
<head>
    <style>
        .bar { color: blue; }
        .foo { color: red; }
    </style>
</head>

<body>
    <p class="foo">Foo</p>
    <p class="bar">Bar</p>
    <p class="foo bar">Foo Bar</p>
    <p class="bar foo">Bar Foo</p>
</body>

</html>
```

The rendered output is:

image

So, my current implementation doesn't work the same as HTML, though in the same spirit of "last wins".

In theory, we could make our implementation work like HTML, but as we don't have the concepts of specificity or the !important modifier, I think this would make it harder for users to understand how to override.

Hence, I think the order of specified classes is the right way to override. The question is whether the left-most one wins or the right-most one does.

I've picked right-most wins, but I'm open to persuasion.

Thoughts?

akrabat:
> * How will this hold up during inheritance? Are things considered non-default if a base class overrides things? 

The concept of the default value for an attribute is defined in ReportLab's `ParagraphStyle` in [styles.py#l112](https://hg.reportlab.com/hg-public/reportlab/file/820380acd8b4/src/reportlab/lib/styles.py#l112) as every style defined in rst2pdf is an instance of `ParagraphStyle` ([rst2pdf/styles.py#L516](https://github.com/rst2pdf/rst2pdf/blob/9e9ce5fca706451dfd0082f2fa3dbfde2d633229/rst2pdf/styles.py#L516)).

An attribute is considered non-default if it has a definition different to what ReportLab sets it to regardless of whether that difference was declared in this class or a parent of it.

When a ParagraphStyle is retrieved from ReportLab's `StyleSheet` collection, the inheritance has already been dealt with and the returned style has all properties set as per the definitions in the chain of parents.

Hence if you have a class called `Foo` that sets `textColor` and has a parent of `Common` which sets `fontSize`, then when we look at `Foo`, both `textColor` and `fontSize` are non-default.

To mildly complicate matters, rst2pdf will set a style's parent to `base`  (defined in styles.yaml) if it doesn't already have a parent declared, so any difference between rst2dpf's `base` and ParagraphStyle's defaults are also considered non-default. This doesn't practically matter though as these particular changes will apply to every defined class, so can be ignored. 

> E.g. both `Foo` and `Bar` have `Base` as their parent. `Foo` has `textColor: red`, `Bar` doesn't have `textColor` specified and `Base` has `textColor: green`. What color will I get if I specify `.. class:: Foo Bar`? Will the inherited green color trump the red? And what if the inheritance is not from a common base?

The right most style's attribute will win, regardless of where in its inheritance chain it was declared. See this test:

image

Relevant files:

[test_style_inheritance.rst.txt](https://github.com/rst2pdf/rst2pdf/files/7740813/test_style_inheritance.rst.txt)
[test_style_inheritance.yaml.txt](https://github.com/rst2pdf/rst2pdf/files/7740814/test_style_inheritance.yaml.txt)
[test_style_inheritance.pdf](https://github.com/rst2pdf/rst2pdf/files/7740817/test_style_inheritance.pdf)

CendioOssman:
> I've picked right-most wins, but I'm open to persuasion.
> 
> Thoughts?

I don't have any strong opinions either way, so I'm fine with the current suggestion.

> The right most style's attribute will win, regardless of where in its inheritance chain it was declared. See this test:

I understand this is probably the easiest to get right from a coding point of view, but I'm worried it might make it more difficult to use style inheritance this way and you're forced to have more redundancy in your style definitions. :/

You can probably avoid this if you make sure all extra classes are just simple modifications, and aren't useful by themselves. Then you basically construct the inheritance in the class list instead. I'm not sure how well that works for Sphinx though where you have little control over the classes assigned.

Have you looked at something like "most specific wins"? Or nested styling like HTML/CSS does where outer elements affect inner ones?

akrabat:
> Have you looked at something like "most specific wins"? Or nested styling like HTML/CSS does where outer elements affect inner ones?

From what I can tell there's no concept of "most specific wins"? The `.. class::`` list occurs as a standalone element in the tree that is then applied to whatever element is discovered next in the tree.

Similarly, a given class just exists with a parent. There is no concept of "p.foo" for instance. 

At least, I can't see any way to do this within the codebase, but could be missing something.

rst2pdf/tests/input/test_multiple_styles.rst
```diff
@@ -0,0 +1,41 @@
+Multiple styles test
```
lornajane, line 1: Hurrah for tests and tests that we can tell from the output what was supposed to happen :) Are there some corner cases where there could be stylesheet styles on an element because of its type, plus a classname? I think more tests could be useful.

lornajane (commented):
Only commenting because I haven't actually put this through its paces yet, and I have questions. But I love the feature and it brings it closer to the functionality people see when outputting to HTML :heart:

Issue #1037, opened 2021-10-28
RFE: Ability to have collapsed outlines


CendioOssman:
**Description of problem**

The outline is currently hard coded to be expanded by default. For large documents this can be problematic as the user then has to scroll a lot to find sections and it is difficult to get an overview. It would be nice to have an option to have it collapsed instead, perhaps even a depth setting.

I'm using Sphinx, so I'd like the option to be exposed via the pdfbuilder as well.

**What is the expected output? What do you see instead?**

I'd like the outline to be collapsed to just the top level headings by default. Currently the entire tree is visible by default.

🖥 **Versions**

* Python 3.6.12
* rst2pdf 0.98
* Reportlab 3.6.1
* Sphinx 4.2.0

akrabat:
I'm not sure what an outline is? Where would I see this? rst2pdf has a table of contents what when rendered in macOS Preview's sidebar, is collapsed:

<img width="219" alt="image" src="https://user-images.githubusercontent.com/33135/145676763-29601f0e-5397-4c3f-8726-1cecddad38d5.png">

Would you be able to provide some further guidance on what you are looking for with examples?

CendioOssman:
That is indeed the outline in your screenshot. It would seem macOS' PDF viewer is buggy. As is Chrome. In a correctly implemented PDF viewer all of those entries are fully expanded. E.g. Firefox or Evince.

The code that does this is here:

https://github.com/rst2pdf/rst2pdf/blob/ae62cb835ad8e312d7366f22d4729ecca06057f7/rst2pdf/flowables.py#L136

The last argument there is if the entry should be close/collapsed. So an option to change that to `True` is what we're requesting.

akrabat:
Thanks, that's clearer. I've added the _help wanted_ label to indicate that if anyone wants to take this on, we'd love to accept the contribution.

akrabat:
> That is indeed the outline in your screenshot. It would seem macOS' PDF viewer is buggy. As is Chrome. In a correctly implemented PDF viewer all of those entries are fully expanded. E.g. Firefox or Evince.
> 
> The code that does this is here:
> 
> https://github.com/rst2pdf/rst2pdf/blob/ae62cb835ad8e312d7366f22d4729ecca06057f7/rst2pdf/flowables.py#L136
> 
> The last argument there is if the entry should be close/collapsed. So an option to change that to `True` is what we're requesting.

How would you imagine this could be set?

akrabat:
[...]

PR #1049, opened 2021-12-19
Collapse outline entries that are deeper than the ToC

akrabat:
Ensure that the level for open outline entries that are usually viewed in the PDF viewer as a collapsible tree matches the number of levels in the table of contents. Any further nested headings will be available in the outline, but collapsed.

Makes #1037 less of a problem.


akrabat:
Rebased to latest master, fixing conflicts.

PR #1062, opened 2022-01-21
Correctly handle '--fit-literal-mode=error' option

mkurmag:
`fit-literal-mode` option sets the behavior when a literal is too wide. There is an `error` value which behaves unexpectedly.

It seems to me that rst2pdf with `--fit-literal-mode=error`should return an error when a literal is too wide and should not produce a PDF.  But currently only warnings are printed and PDF is successfully created.

PR fixes behavior and adds a corresponding test.


akrabat (approved):
Thanks. This is great. I'm sorry that it took so long.

PR #1081, opened 2022-09-18
Drop Python 3.6, update dependencies

lornajane:
Python 3.6 is end of life, and by dropping it we can go back to one `requirements.txt` file.

Also we have some outstanding dependabot updates that aren't captured by setup.py plus pip compile, so this PR also adds an instruction to dependabot on how to update our dependencies list.


akrabat:
I have removed 3.6 from the list of required checks in the branch protection rules.

akrabat (approved):

PR #1087, opened 2022-09-19
Move tests directory to root

akrabat:
This is something I've been meaning to do for ages as the `tests` directory doesn't belong in the `rst2pdf` directory.


lornajane (approved):
This makes perfect sense, thanks (sorry for the slow review!)

Issue #1080, opened 2022-09-15
Warnings: [CRITICAL] math_flowable.py:169 descent = 12.0;  [CRITICAL] math_flowable.py:174 chr(num) = N;


arigalzi:
**Description of problem**

I am using Sphinx and when including math formulas a warning appears during the build of the documents for each char inside it.
I am using the sphinx extension to process them ([sphinx.ext.mathjax](https://sphinx-rtd-trial.readthedocs.io/en/latest/ext/math.html#module-sphinx.ext.mathjax)) and an example line appears to be like this:

r"""

:math:`rel_i(r)= 1`
 
"""

which causes when creating the pdf docs to output in console:

[CRITICAL] math_flowable.py:169 descent = 2.0
[CRITICAL] math_flowable.py:174 chr(num) = r
[CRITICAL] math_flowable.py:174 chr(num) = e
[CRITICAL] math_flowable.py:174 chr(num) = l
[CRITICAL] math_flowable.py:174 chr(num) = i
[CRITICAL] math_flowable.py:174 chr(num) = (
[CRITICAL] math_flowable.py:174 chr(num) = r
[CRITICAL] math_flowable.py:174 chr(num) = )
[CRITICAL] math_flowable.py:174 chr(num) = =
[CRITICAL] math_flowable.py:174 chr(num) = 1

Generated from this log.critical() in math_flowable.py script https://github.com/rst2pdf/rst2pdf/blob/b502a2bc66da178ee0a30e91155d4ac455c0fc85/rst2pdf/math_flowable.py#L174

**What is the expected output? What do you see instead?**
The math formulas correctly appear in the pdf and I couldn't find any further details on how to avoid these warnings. Either adding a description on the reason of the critical warnings or deleting them if not necessary would be really useful.

🖥 **Versions**

`python -V`
Python 3.8.13

`pip freeze | grep rst2pdf`
rst2pdf==0.99

`pip freeze | grep reportlab`
reportlab==3.6.11

`pip freeze | grep Sphinx`
Sphinx==5.1.1

Which operating system are you using?
Linux

lornajane:
The math support in rst2pdf is not as complete as in docutils. I think this is a valid bug, so if anyone can fix it, we'd happily review a patch.

PR #1090, opened 2022-10-17
math_flowable: remove unnecessary critical warining

arigalzi:
A critical warning appears during the build of the documents for each char inside the math formula even if it correctly appears in the pdf since the math support is not yet complete in rst2pdf,. For this reason, the log.critical() lines that generate them are being remove from math_flowable.py.

Signed-off-by: Arianna Galzerano <ariannagalzerano@gmail.com>
Closes: #1080


akrabat:
Rebased against latest main.

akrabat (approved):
Thanks for this!

Issue #1067, opened 2022-04-08
0.99: 16 failed tests


dvzrv:
Hi! I'm packaging rst2pdf for Arch Linux.
In #1015 I tried to clear up the required dependencies of this project. As mentioned in https://github.com/rst2pdf/rst2pdf/issues/1015#issuecomment-1027187929 I am still met with 16 failing tests though.

**Description of problem**

```
python -m build --wheel --no-isolation
local _site_packages=$(python -c "import site; print(site.getsitepackages()[0])")


mkdir -vp test_dir
python -m installer --destdir=test_dir dist/*.whl

export PYTHONPATH="$PWD/test_dir/$_site_packages:$PYTHONPATH"
export PATH="$PWD/$_test_dir/usr/bin:$PATH"
pytest -vv
```

leads to:

```
=========================== short test summary info ============================
FAILED rst2pdf/tests/input/test_box_widths.rst::test_box_widths
FAILED rst2pdf/tests/input/test_header_footer_in_stylesheet.rst::test_header_footer_in_stylesheet
FAILED rst2pdf/tests/input/test_issue_110.rst::test_issue_110
FAILED rst2pdf/tests/input/test_issue_110_2.rst::test_issue_110_2
FAILED rst2pdf/tests/input/test_issue_110_3.rst::test_issue_110_3
FAILED rst2pdf/tests/input/test_issue_126.rst::test_issue_126
FAILED rst2pdf/tests/input/test_issue_261.rst::test_issue_261
FAILED rst2pdf/tests/input/test_issue_274_2.rst::test_issue_274_2
FAILED rst2pdf/tests/input/test_issue_288.rst::test_issue_288
FAILED rst2pdf/tests/input/test_issue_295.rst::test_issue_295
FAILED rst2pdf/tests/input/test_preprocessor.rst::test_preprocessor
FAILED rst2pdf/tests/input/test_raw_html.rst::test_raw_html
FAILED rst2pdf/tests/input/test_stylesheet_includes.rst::test_stylesheet_includes
FAILED rst2pdf/tests/input/test_table_in_header.rst::test_table_in_header
FAILED rst2pdf/tests/input/test_uml_extension.rst::test_uml_extension
FAILED rst2pdf/tests/input/sphinx-footnotes-order/conf.py::sphinx-footnotes-order
============ 16 failed, 275 passed, 22 skipped in 152.36s (0:02:32) ============
```

Full build log: 
[...]

PR #1096, opened 2022-11-15
Recompile requirements since we have a few dependabot patches open

lornajane:
Updated requirements list from setup.py, since we have a few dependencies that should be patched.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
     continue-on-error: ${{ matrix.experimental }}
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9, '3.10']
+        python-version: ['3.8', '3.9', '3.10']
         experimental: [true]
 
     steps:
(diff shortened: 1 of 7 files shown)


lornajane:
This does include everything that dependabot has suggested with its open PRs.

It excludes Reportlab - 3.6.3 is what we were using, the next release is 3.6.5 and it changes spacing in a footer and causes a handful of tests to fail, I think #1067  is the same tests failing. We need to figure out if we can do something with this, or relax the tests a bit. This is the only departure from the compiled dependencies output.

There is a bonus fix to the pre-commit configuration as it seems like flake8 moved its main repo.

I've switched a switch on the github workflows so that we do see all the versions run if one fails, we probably should back that out before we merge.

Python 3.7 does not seem to be able to resolve these dependencies - it's security fixes only so we need to decide if we're prepared to drop support for it.

akrabat:
So we have two choices in terms of supported Python version:

1. Support the same list of versions python.org. That's 3.7 onwards until June 2023 when 3.7 reaches end of life.
2. Support the minimum version that our pinned dependencies support. That's 3.8 due to matplotlib 3.6.x.

For 1, we would need to go back to having multiple requirements.txt for different Python versions.

I have no strong opinion either way.

lornajane:
I vote option 2 - 3.7 is fairly close to end of life so people should use an older revision of rst2pdf if they're on 3.7.

PR #1118, opened 2023-01-11
Add config option to change default subtitle prefix

matthew-ivre:
I think that the end user should have the option to change the default subtitle prefix. For example, from "version" to "release" or "v.". It would also allow the end user to set the prefix to a empty string to remove it completely when not needed.


lornajane:
Thanks for the pull request and sorry for the slow response on this - the builds didn't run and it took me some time to debug them. I've pushed a tiny fix so this patch is ready to review IMO

akrabat:
Rebased to pick up CI updates.

lornajane (approved):
Thanks @matthew-ivre - that looks like a useful addition!

PR #1126, opened 2023-03-08
Restore twoColumn template to the default set

### LINKED ISSUE #1122: sphinx  index directive fails needing twoColumn
**See later comments as the root cause was figured out**

**Description of problem**

Follow the instructions for sphinx integration, and then try to build pdf.

You will get an error:

    Template 'twoColumn' is not defined

This is as expected because that template has been removed from rst2pdf.  The sample config has:

    pdf_page_template = 'cutePage'

However there is no such template in Sphinx or rst2pdf.  The only templates I can find is for the cover page.  This leaves me stuck because I have no idea what should be in the template, and neither project is providing a default.

🖥 **Versions**

`python -V`
Python 3.10.7

`pip freeze | grep rst2pdf`
rst2pdf==0.99

`pip freeze | grep reportlab`
reportlab==3.6.12

`pip freeze | grep Sphinx`
Sphinx==6.1.3

Ubuntu 22.10

**Additional information**

It would be great if rst2pdf worked out of the box with Sphinx.

akrabat:
We use twoColumn when creating Sphinx indexes, so we need it as a default template so that Sphinx projects build.

Fixes #1122


lornajane (approved):
Thanks - this does fix the problem although I wish the tests had picked it up! I think we can test this, but that we shouldn't delay merging this change, so #1129 captures the need for followup

PR #1110, opened 2023-01-07
Update ReportLab to 3.6.12

akrabat:
ReportLab version 3.6.5 changed something about spacing, particular for headers/footers that affected a number of tests. As I can't see anything we can directly do in our code to fix this, I've just updated the tests to accept the new rendering.

PR #1133, opened 2023-03-16
Only allow Python 3.11 builds to fail

akrabat:
We don't want to allow everything to `continue-on-error` as we don't spot when there's a real failure.

We let 3.11 fail as ReportLab 3.6.3 doesn't install on 3.11 and we need move to a more recent ReportLab for that one (#1110).


akrabat:
To be rebased against main when #1134 is merged.

akrabat:
Rebased now that #1134 is merged.

lornajane (approved):
Keeping up with the pythons is great, thank you!

PR #1111, opened 2023-01-08
Write dependency file

t-8ch:
Add new parameter --record-dependencies which works the same way was for vanilla docutils rst2foo commands.

The target file will contain a list of files that were used to generate the output file. The list can be consumed by build-systems to efficiently regenerate the PDF if any of the inputs has changed.

Fixes #1108

PR #1134, opened 2023-03-16
Fix test failures

akrabat:
Due to the way our CI was set up and a failure to manually run the pytest tests locally before merging, we ended up merging two PRS (#1111 and #1126) that broke the tests. This PR fixes both.

For #1126, `test_stylesheet_includes` is now a two-column PDF again so we need to update the reference file.

For #1111, recording dependencies breaks Sphinx, which does't even allow this option to be set. Hence, if the `--record-dependencies` option is not set, we no longer do the recording as we're not going to use the result anyway.


lornajane (approved):
Thanks! Looks good both here in CI and locally

Issue #1146, opened 2023-04-08
Wrong order of arguments in call to SplitTable and DelayedTable


KarstenSchulz:
It seems, that calls to SplitTable and DelayedTable in `genelements.py` have a wrong order of arguments.

The definition of [SplitTable](https://github.com/rst2pdf/rst2pdf/blob/43df27f6a0d10927a82a1e4d6480788ee0a79687/rst2pdf/flowables.py#L339) is:

```
class SplitTable(DelayedTable):
    def __init__(self, data, colWidths, style, padding=3):
    	...
```

The call in [genelements.py](https://github.com/rst2pdf/rst2pdf/blob/43df27f6a0d10927a82a1e4d6480788ee0a79687/rst2pdf/genelements.py#L1076) is:

```
SplitTable([['', rows]], style=t_style, colWidths=[0, None], padding=st.borderPadding,),

```

The order of `style` and `colWidths` are swapped.

The same observation can be made with [DelayedTable](https://github.com/rst2pdf/rst2pdf/blob/43df27f6a0d10927a82a1e4d6480788ee0a79687/rst2pdf/flowables.py#L226). In [line 1092 in the file genelements.py](https://github.com/rst2pdf/rst2pdf/blob/43df27f6a0d10927a82a1e4d6480788ee0a79687/rst2pdf/genelements.py#L1092) are the arguments `style` and `colWidths` swapped.

**Description of problem**

There is no observable problem, all tests pass.

🖥 **Versions**

`python -V` -> `Python 3.10.9`

`pip freeze | grep rst2pdf` -> I use the latest from github

`pip freeze | grep reportlab` -> `reportlab==3.6.12`

Which operating system are you using? MacOS 13.3.1

**Additional information**

If my observation is correct, I will search for other bugs of this kind and prepare a pull request. Please let me know.

akrabat:
As I understand keyword arguments, the order of the keyword arguments when calling a function does not matter. Having said that, consistency makes sense, so I'm okay with matching the order when we call with the defined order.

KarstenSchulz:
Yes, you are right, of course. However, if you want to learn and understand the source code of rst2pdf, as I do, it costs unnecessary mental power if the arguments are listed in a different order than the parameters. I am glad you agree and will finish a pull request in the next days. Thank you.

PR #1150, opened 2023-04-11
Reorder caller args as discussed in #1146

KarstenSchulz:
This PR reorders the caller arguments according to the parameter list of the class to minimize mental debt when studying the code (see #1146). No functional change. Tests pass.


akrabat (approved):
LGTM. Thanks!

PR #1160, opened 2023-05-06
Restore twocolumn.yaml

### LINKED ISSUE #1159: Stylesheet `twocolumn` is not found
**Description of problem**
We recently added the twocolumn stylesheet back into the main codebase #1126. However I don't seem to be able to run `rst2pdf -s twcolumn myfile.rst` - the stylesheet isn't found and my document isn't in two columns.

I notice that the test `test_stylesheet_includes` passes, but it emits a warning into its log and looking at the PDF, it's not in two columns (the style does what it sounds like it would).

I worked around my problem by grabbing the old `twocolumn.yaml` file from the git history and using that locally.

🖥 **Versions**

`python -V`
Python 3.10.8

`pip freeze | grep rst2pdf`
`-e git+ssh://git@github.com/rst2pdf/rst2pdf.git@9bb7e8ee8b597a7cd40f36e89fc10f76c0d0a5ca#egg=rst2pdf` <-- what even is that?

`pip freeze | grep reportlab`
reportlab==3.6.12

Fedora

akrabat:
This file is required as a way to set the two column layout via a stylesheet. It was incorrectly removed in #1126.

Fixes #1159

### DIFF
--- a/rst2pdf/styles/twocolumn.yaml
+++ b/rst2pdf/styles/twocolumn.yaml
@@ -1,0 +1,4 @@
+pageSetup:
+  firstTemplate: twoColumn
+  margin-left: 1cm
+  margin-right: 1cm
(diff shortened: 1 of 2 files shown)


lornajane (approved):
Thank you! This does what I expected :)

PR #1162, opened 2023-05-19
Make the hl_lines directive work with plain text

akrabat:
As plain text code blocks have a different path through the codeblock directive, we need to apply the hl_lines "`diml`" class in this section too in order for highlighting to work.

I've also improved the Python test to show that syntax highlighting works on the highlighted lines and updated an error message to show the name of a class rather than its index within the list of classes.

This is an example from the new test in this PR:

RST code:
```rst
.. code-block:: text
   :linenos:
   :hl_lines: 2

   To be
   or not to be
   that is the question
```

Before this PR, only the first line number is dimmed:
image

After, it works as you'd reasonably expect:
image


lornajane (approved):
This looks great, thank you :)

PR #1166, opened 2023-05-24
Update installation instructions for pipx (and other tidy-ups)

### LINKED ISSUE #1165: Installation instructions need updating for (at least) Debian/Ubuntu
Newer versions of Debian and related Linux distros actively discourage global installation of Python packages using `pip`.  Which makes sense unless you're trying to install a commandline utility application such as ... rst2pdf. The user will see a message like:

```
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.
    
    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.
    
    See /usr/share/doc/python3.11/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
```

This doesn't make much sense for a tool like rst2pdf which is likely to be wanted in lots of different projects - it's a commandline utility rather than a project-specific library.

We have a few options:

* Users can work around the problem with `PIP_REQUIRE_VIRTUALENV=false pip install rst2pdf` or `pip install rst2pdf --break-system-packages`
* It should be possible to install (possibly older) versions of `rst2pdf` using `apt`
* Update the instructions to support installing through [pipx](https://pypa.github.io/pipx/) which isolates dependencies to the application.

We should probably do the third option anyway at some point, it's more pythonic-future. Let's leave this issue open until we have pipx installation and instructions.

akrabat:
With the adoption of PEP 668, installation of applications such as rst2pdf with pip is discouraged and pipx is a better choice. Update our instructions to use pipx. 

I also removed the Travis badge and renamed `master` to `main` in the README.

Finally, I've improved DEVELOPERS.rst to provide more comprehensive instructions for setting up rst2pdf for development with a venv and tidied up some other bits so that they are
more accurate/useful.

Fixes #1165

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,6 +1,3 @@
-.. image:: https://travis-ci.org/rst2pdf/rst2pdf.svg?branch=master
-    :target: https://travis-ci.org/rst2pdf/rst2pdf
-
 .. image:: https://img.shields.io/pypi/v/rst2pdf.svg
     :target: https://pypi.org/project/rst2pdf/
 
@@ -83,7 +80,7 @@ can clone the repository and install this version::
 
     $ git clone https://github.com/rst2pdf/rst2pdf
     $ cd rst2pdf
-    $ git checkout <desired-branch> # if you want something other than master
+    $ git checkout <desired-branch> # if you want something other than main
     $ pipx install .[aafiguresupport,mathsupport,plantumlsupport,rawhtmlsupport,sphinx,svgsupport]
 
 If you intend to work on rst2pdf's source code, see ``doc/DEVELOPERS.rst``.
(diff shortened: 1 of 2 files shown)


lornajane (approved):
Thanks! I think you just about managed a patch longer than my original epic bug report!!

PR #1167, opened 2023-05-31
add pythonpublish.yml

### LINKED ISSUE #1154: Publish to PyPI using a GH action
We should automate creating the PyPI release using https://github.com/pypa/gh-action-pypi-publish. We could probably do this on push of a version tag, where we automatically create the GH release to, or maybe we just do upon creation of the release.

sigma67:
Note: This requires a GitHub secret PYPI_API_TOKEN, which is up to the maintainers to configure.

This publishes to PyPi whenever a release is created, based on the Git tag associated with that release.

Closes #1154

### DIFF
--- a/.github/workflows/pythonpublish.yml
+++ b/.github/workflows/pythonpublish.yml
@@ -1,0 +1,30 @@
+name: Upload Python Package
+
+on:
+  release:
+    types: [published]
+
+permissions:
+  contents: read
+
+jobs:
+  deploy:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python
+        uses: actions/setup-python@v3
+        with:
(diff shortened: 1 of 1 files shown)


FriedrichFroebel:
Just as a heads-up: PyPI nowadays recommends using Trusted Publishers instead of token-based authentication: https://docs.pypi.org/trusted-publishers/using-a-publisher/

sigma67:
Yes, there have been changes in the meantime. I can update it, but I'm not sure there's interest from the maintainers since none of my PRs got through

lornajane:
Thanks @FriedrichFroebel, the extra information is always appreciated.

@sigma67 the maintainers are volunteers with a large number of other commitments and limited Python expertise. We do appreciate your input in the project, but I do not find your remark about a lack of interest either constructive or helpful.

PR #1201, opened 2024-04-03
pdfbuilder: Add supported_image_types attribute

castler:
According to the Sphinx documentation Builders should include an attribute `supported_image_types`, that lists which images types can be used by this builder.

https://www.sphinx-doc.org/en/master/extdev/builderapi.html#sphinx.builders.Builder

Other plugins rely on this behavior.

By specifying the list of image types, we ensure that the API exposes the capabilities correclty (since rst2pdf has image support).


lornajane (approved):
Thanks for contributing this, it's a good addition!

(it looks like GitHub didn't run the checks, but the test suite is fine for me locally so I'm calling this good)

PR #1214, opened 2024-05-27
Support ReportLab 4 & xhtm2pdf 0.12.2+

akrabat:
Update requirements.txt to latest version of all dependencies, in particular picking up ReportLab 4.

A number of tests failed, which I reviewed and determined that the changes to rendering were either an improvement or an acceptable difference.

I also updated the signature for `pytest_collect_file` which has been updated upstream to use a PathLib object for the file path.


akrabat:
Note that Python 3.8 support is removed due to dependencies no longer supporting it. This means that the build requirement of a passing 3.8 test will never happen.

lornajane (approved):
Thanks for the updates!

PR #1217, opened 2024-06-05
Update RELEASE_PROCESS.rst

akrabat:
Various updates after releasing a new version to hopefully make it clearer next time.

### DIFF
--- a/doc/RELEASE_PROCESS.rst
+++ b/doc/RELEASE_PROCESS.rst
@@ -9,7 +9,7 @@ This is an outline of what needs to be done in order to release rst2pdf.
 
       $ pip install setuptools setuptools_scm wheel twine
 
-#. Update ``CHANGES.rst`` to add the version number and date. Commit to a branch, PR and merge to master
+#. Update ``CHANGES.rst`` to add the version number and date. Commit to a branch, PR and merge to main
 #. Ensure all PRs are attached to the milestone
 #. Close the milestone and create next one
 #. Checkout main and ensure that your working copy is clean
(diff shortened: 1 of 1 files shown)


doc/RELEASE_PROCESS.rst
```diff
@@ -33,39 +33,72 @@ This is an outline of what needs to be done in order to release rst2pdf.
 
 #. Build manual
 
-   You will need the `White Rabbit`_ font installed. We check out the tag first so that the version
+   You will need the `White Rabbit`_ font installed. We check out the tag install via pip first so that the version
    on the first page of the PDF is labelled as final and then generate the HTMl and PDF docs.
```
lornajane, line 37:
```suggestion
   on the first page of the PDF is labelled as final and then generate the HTML and PDF docs.
```

doc/RELEASE_PROCESS.rst
```diff
@@ -33,39 +33,72 @@ This is an outline of what needs to be done in order to release rst2pdf.
 
 #. Build manual
 
-   You will need the `White Rabbit`_ font installed. We check out the tag first so that the version
+   You will need the `White Rabbit`_ font installed. We check out the tag install via pip first so that the version
```
lornajane, line 36: I'm not sure this works. Is it "check out the tag AND install via pip" ? I think checking out the tag first links to the second half of the sentence which is why I'm confused.

doc/RELEASE_PROCESS.rst
```diff
@@ -33,39 +33,72 @@ This is an outline of what needs to be done in order to release rst2pdf.
...
+    Double check that you do not have any changes in your working directory that are not committed. If you do, stash
+    them as otherwise uploading to PyPI will not work.
+
     ::
 
        $ python setup.py egg_info -b "rc1" sdist bdist_wheel
 
     If you're doing an alphaX, betaX or postX, then change ``-b "rc1"`` appropriately
 
+#. Setup PyPI if you haven't already
```
lornajane, line 75:
```suggestion
#. Set up PyPI if you haven't already
```

lornajane (commented):
I'm sure this is am improvement but there's one bit that tripped me up, can you help to untangle the meaning(s)?

doc/RELEASE_PROCESS.rst
akrabat: You're right. This is not good. I've taken another stab at it, including moving the font requirement to the top as that's where we mention the other dependencies required.

lornajane (approved):
Thanks for updating this for future-us :)

Issue #738, opened 2019-02-11
Rewrite setup.py / requirements


ralsina:
It's currently very old fashioned (as is to be expected ;-)

It should be redone using something like [poetry](https://github.com/sdispater/poetry) after the py3 port

kyluca:
Hi,

Just wanting to check if this is still desirable to the project? If so, I'd be happy to put up a PR for Hacktoberfest?

Poetry was a great suggestion in 2019, but now there's an official guide to "modernize" a setuptools project with first class support in `pyproject.toml`: https://packaging.python.org/en/latest/guides/modernize-setup-py-project/

Cheers,
Kyle

lornajane:
@kyluca yes please, but I think this issue is missing some of the context.

Right now we're using pip-tools with setup.py. I also notice some tox.ini and other files that I don't think are still in use.
We publish the package for others to consume, so our sdist output is also important.

The maintainers on this project aren't Python experts, so we will also need instructions on how to update our existing workflows to use the new approach - we won't already know what to do! If that sounds good, let me know and I'll tag/assign the issue for you.

kyluca:
> The maintainers on this project aren't Python experts, so we will also need instructions on how to update our existing workflows to use the new approach - we won't already know what to do!

No worries!

`pip-tools` is also a great tool (as was `poetry` for its time), although my two-cents on the leading solution for modern projects would be to use `uv`: https://github.com/astral-sh/uv

Even though `uv` is a drop-in replacement for `pip-tools`, it might be worth making this change in two steps to minimise disruption. I can do one PR focused on migrating `setup.py` to `pyproject.toml` while keeping the current usage of `pip-tools`. Then a follow up PR to illustrate how one might adopt `uv` in the place of `pip-tools`. The maintainers can then choose if they're happy to go with the full change or just the first half 👍🏾 

If that's sensible, I'll try to get the first PR up over the next few days 😄

lornajane:
That does sound sensible, thank you!

kyluca:
PR up for the first half, let me know what you think :)

I haven't removed the `tox.ini` yet, just wanted to start with the minimal changes to get `pyproject.toml` fully up and running. Can include the removal (or revival?) of `tox` in the follow up PR.

PR #1241, opened 2024-10-19
Fix sphinx test collection logic

kyluca:
While working on my second PR for #738, I noticed that the sample test command in `doc/DEVELOPERS.rst` no longer works:

> Running a single test
> *********************
> 
> To run one test only, simply pass the file or directory to pytest. For example::
> 
>   pytest tests/input/sphinx-repeat-table-rows

Local result:

```
 $  uv run pytest tests/input/sphinx-repeat-table-rows
================================ test session starts =================================
platform linux -- Python 3.10.12, pytest-8.3.3, pluggy-1.5.0
rootdir: /home/kyle/repos/rst2pdf
configfile: pyproject.toml
plugins: xdist-3.6.1
collected 0 items

=============================== no tests ran in 0.04s ================================
```

After some digging, I found it was because the custom test collection logic used `path.stem` (e.g. `conf`) instead of `path.name` (e.g. `conf.py`), but was expecting to compare it to `conf.py`.

Before the fix, there are 275 tests collected (matches current CI runs on GHA):

```
 $  uv run pytest --collect-only
================================ test session starts =================================
platform linux -- Python 3.10.12, pytest-8.3.3, pluggy-1.5.0
rootdir: /home/kyle/repos/rst2pdf
configfile: pyproject.toml
plugins: xdist-3.6.1
collected 275 items

...

============================ 275 tests collected in 0.13s ============================
```

After the fix, there are 315 tests collected (all of those starting with `sphinx`):

```
 $  uv run pytest --collect-only
================================ test session starts =================================
platform linux -- Python 3.10.12, pytest-8.3.3, pluggy-1.5.0
rootdir: /home/kyle/repos/rst2pdf
configfile: pyproject.toml
plugins: xdist-3.6.1
collected 315 items

...

============================ 315 tests collected in 0.13s ============================
```

Bug appeared to be accidentally introduced in 3ad61be8d

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -432,7 +432,7 @@ def pytest_collect_file(file_path, parent):
 
     if file_path.suffix == '.rst' and parent_dir == 'input':
         return RstFile.from_parent(parent=parent, path=file_path)
-    elif file_path.stem == 'conf.py' and parent_dir.startswith('sphinx'):
+    elif file_path.name == 'conf.py' and parent_dir.startswith('sphinx'):
         return SphinxFile.from_parent(parent=parent, path=file_path)
 
 


akrabat:
This is a bit more complicated than that as the Sphinx tests create a directory that contains the output PDF to be compared.

akrabat:
Why we add an additional document is most definitely lost in the mist of time! Maybe @ralsina remembers?

akrabat:
I can't think of any reason to not get rid of that additional doc as we could document in the manual how to add it back in `conf.py` and let the user explicitly include if if they need it. 

We could also remove `master_doc` from all the tests too as it doesn't seem to be documented in the manual anway.

lornajane:
Sphinx has something different about its cover page I think

lornajane:
First of all: I think this PR is correct and useful, and fixes a real bug. Thanks @kyluca ! However it does expose a bunch of other things that we will now need to fix ..... (this is absolutely a Good Thing (TM) )

I think this block of code was there as a sort of default set of PDF document properties in case nothing was specified - for example two tests `sphinx-issue319` and `sphinx-issue320` are missing the `pdf_documents` entry from their `conf.py` files so without this block present, no PDF is created at all!

I'm pushing a branch to try to fix/mitigate some of the test problems so we can unblock you and your other work --> it's in #1243, includes the changes in this pull request as a basis, but still needs a little bit more to get all the tests to pass.

kyluca:
Right, that makes sense. #1243 looks good, although I'm in no particular rush.

Let me know if there's anything else I can do 👍🏾

PR #1240, opened 2024-10-11
Migrate setup.py to pyproject.toml

kyluca:
Implements the first half of the migration as discussed on #738.

Essentially followed the official advice from https://packaging.python.org/en/latest/guides/modernize-setup-py-project/. Every field from the old `setup.py` config has an analogue in `pyproject.toml` just with slightly different syntax.

I've also confirmed that the project can still be built into both a source distribution (sdist) and a wheel (bdist). Corresponding documentation has been updated from using `python setup.py ... sdist bdist_wheel` to the newer `python -m build` as the file `setup.py` itself is no longer required.

This is a screenshot of the contents of the wheel of the latest release downloaded from PyPI on the left hand side vs. the contents of a locally built wheel for a 0.103 rc1 package showing the non `*.py` package data files are still present.

screenshot_from_2024_10_12_00_17_43

I've also verified the unit test suite still passes locally after all the changes (in particular to how things like package data is handled).

Let me know if there's anything else I can update at this stage. If this looks okay I'll continue with a follow up PR on consolidating all of the `pip`, `pip-tools`, `pipx` and `venv` usage to just `uv` (second half of the plan as discussed on the issue). Although that second set of changes is not required, it's just a "nice to have".

PR #1242, opened 2024-10-21
 Adopt uv as a replacement for pip, pipx, pip-tools and virtualenv

kyluca:
As discussed in #738 and to follow on from the migration in #1240, now that the old `setup.py` has been modernised to `pyproject.toml`, we can take advantage of newer dev tooling.

As per their docs, [uv](https://github.com/astral-sh/uv) is:

> 🚀 A single tool to replace pip, pip-tools, pipx, poetry, pyenv, virtualenv, and more.
⚡️ [10-100x faster](https://github.com/astral-sh/uv/blob/main/BENCHMARKS.md) than pip.

In this PR I've replaced all usages of `pip`, `pipx`, `pip-tools` (`pip-sync` and `pip-compile`), `build`, `twine` and `virtualenv` management with the single tool `uv`.

My personal experience with it so far has been really smooth, but I appreciate the nature of this change may feel disruptive at first. Please take your time to read, ask questions, checkout the branch and try some of the commands.

Also note I've based this branch off #1241, so that PR should be reviewed and merged independently first.

There's also one minor bugfix included here which I came across while updating the GHA workflow, in that the `lint` stage has not been running since the CI suite moved off Python 3.8. Updated now to run during the 3.9 build.

Let me know what you think!

Cheers,
Kyle


.github/workflows/ci.yml
```diff
@@ -17,28 +17,27 @@ jobs:
...
-        pip install pytest pre-commit
     - name: Install dependencies for Python
       run: |
-        pip install -c requirements.txt -e .[tests,sphinx,svgsupport,aafiguresupport,mathsupport,rawhtmlsupport]
+        uv sync --all-extras --python ${{ matrix.python-version }}
     - name: Lint
       run: |
-        pre-commit run --all-files --show-diff-on-failure
-      if: matrix.python-version == '3.8'
+        uvx pre-commit run --all-files --show-diff-on-failure
```
kyluca, line 36: `uvx` is a convenience helper to install and run arbitrary tools in isolated environments on-demand. In practice this means that we can use `pre-commit run` in this one liner without needing to first install it like the old code did on L31.

.github/workflows/ci.yml
```diff
@@ -17,28 +17,27 @@ jobs:
...
-        pip install --upgrade setuptools
-        pip install pytest pre-commit
     - name: Install dependencies for Python
       run: |
-        pip install -c requirements.txt -e .[tests,sphinx,svgsupport,aafiguresupport,mathsupport,rawhtmlsupport]
+        uv sync --all-extras --python ${{ matrix.python-version }}
     - name: Lint
       run: |
-        pre-commit run --all-files --show-diff-on-failure
-      if: matrix.python-version == '3.8'
```
kyluca, line 38: This was the bugfix mentioned in the PR description. Python 3.8 isn't present in the testing version matrix, so it was never executed (see previous CI runs on `main`).

CONTRIBUTING.rst
```diff
@@ -104,31 +106,30 @@ Set up to run tests
 *******************
 
 To run the tests for the first time, you will need to do some setup (after
-this, you can just work on your given virtualenv each time)::
+this, any ``uv`` commands will automatically work within the right virtual
+environment)::
 
-    python -m venv .venv
-    . .venv/bin/activate
+    uv sync --all-extras
```
kyluca, line 112: This shows the simplicity of using `uv` instead of managing virtualenvs by hand. It's also extremely fast compared to the old `pip install` commands (please try this out locally!)

CONTRIBUTING.rst
```diff
@@ -104,31 +106,30 @@ Set up to run tests
...
-    pip install -c requirements.txt -e .[aafiguresupport,mathsupport,plantumlsupport,rawhtmlsupport,sphinx,svgsupport,tests]
+If you don't have ``uv``, please see `the installation docs <https://docs.astral.sh/uv/getting-started/installation/>`_
 
 Run tests
 *********
 
 To run all tests, run::
 
-  pytest
+    uv run pytest
```
kyluca, line 121: Instead of having `pytest` in your global path (inside the virtualenv), prefixing commands with `uv run` ensures they're running within the context of your project's environment without any ambiguity.

README.rst
```diff
@@ -49,19 +49,22 @@ Features
...
+last version to support Python 2.7.
 
 Install from PyPI
 ~~~~~~~~~~~~~~~~~
 
-The latest released version may be installed from PyPI by using ``pipx``::
+The latest released version may be installed from PyPI by using `uv <https://docs.astral.sh/uv>`_::
 
-    $ pipx install rst2pdf
+    $ uv tool install rst2pdf
```
kyluca, line 60: Here `uv tool` is a replacement for `pipx`, but I note the absence of `pip` instructions for downstream users who may not be using `pipx` or `uv` in their own workflows.

Would it make sense to add `pip` instructions below as another alternative method?

doc/DEVELOPERS.rst
```diff
@@ -36,8 +36,8 @@ If you want to do something inside rst2pdf, you are welcome! The process looks s
 * If you added a command line option, document it in ``doc/rst2pdf.txt``.  That
   will make it appear in the manual and in the man page.
 
-* If you implemented a new feature, please document it in ``manual.rst`` (or in
-  a separate file and add an include in ``manual.rst``)
+* If you implemented a new feature, please document it in ``doc/manual.rst`` (or in
+  a separate file and add an include in ``doc/manual.rst``)
```
kyluca, line 40: Minor docs fix.

doc/DEVELOPERS.rst
```diff
@@ -50,25 +50,11 @@ activate a venv::
 
     git clone https://github.com/rst2pdf/rst2pdf
     cd rst2pdf
-    python3 -m venv .venv
-    . .venv/bin/activate
+    uv sync --all-extras
```
kyluca, line 53: As above, this shows the power of `uv` compared to the old setup process on the LHS.

doc/DEVELOPERS.rst
```diff
@@ -50,25 +50,11 @@ activate a venv::
...
 
-    pip install --upgrade setuptools pip
-
-Now you can install rst2pdf from this source code::
-
-    pip install -c requirements.txt -e .[aafiguresupport,mathsupport,plantumlsupport,rawhtmlsupport,sphinx,svgsupport,tests]
-
-Note, that on Apple Silicon Macs, you may need this to get tests passing::
-
-    pip install reportlab==3.6.12 --force-reinstall --no-cache-dir --global-option=build_ext
```
kyluca, line 66: Note this special case should no longer be required after `reportlab>=4.0.0` which is pure Python according to their [release notes](https://docs.reportlab.com/install/open_source_installation/):

> Since ReportLab 4.0 we have moved to pure Python!

doc/DEVELOPERS.rst
```diff
@@ -121,26 +106,20 @@ caused by these changes in underlying dependencies.
...
+To run all the tests via ``pytest`` use::
 
-    pytest
+    uv run pytest
 
 You can also run tests in parallel using ``pytest-xdist`` by passing the ``-n auto`` flag.
 
-Firstly install::
-
-    pip install pytest-xdist
```
kyluca, line 132: Moved `pytest-xdist` into the dev dependencies defined in `pyproject.toml` rather than just a suggestion in the docs.

doc/RELEASE_PROCESS.rst
```diff
@@ -4,11 +4,7 @@ Release Process for rst2pdf
 
 This is an outline of what needs to be done in order to release rst2pdf.
 
-#. Install dependencies that you'll need
-   ::
-
-      $ pip install build twine
```
kyluca, line 10: No need for additional disparate tools as `uv build` replaces `python -m build` and `uv publish` replaces `twine`

doc/RELEASE_PROCESS.rst
```diff
@@ -78,69 +74,47 @@ This is an outline of what needs to be done in order to release rst2pdf.
...
        $ echo -e "[egg_info]\ntag_build=rc1\n" > /tmp/build_opts.cfg
-       $ DIST_EXTRA_CONFIG=/tmp/build_opts.cfg python -m build
+       $ DIST_EXTRA_CONFIG=/tmp/build_opts.cfg uv build
        $ rm /tmp/build_opts.cfg
 
     If you're doing an alphaX, betaX or postX, then change ``tag_build=rc1`` appropriately
 
-#. Set up PyPI if you haven't already
-
-    Create a ``~/.pypirc`` file with sections for rst2pdf and testrstpdf which are then used with ``twine`` via the
```
kyluca, line 88: `uv publish` doesn't use `~/.pypirc` like `twine` did previously, so we no longer require this step.

doc/RELEASE_PROCESS.rst
```diff
@@ -78,69 +74,47 @@ This is an outline of what needs to be done in order to release rst2pdf.
...
+   rc release on Test PyPI and be able to create PDF documents from rst files e.g.
 
     ::
 
-       $ pip install --index-url https://test.pypi.org/simple \
-         --extra-index-url https://pypi.org/simple --pre rst2pdf
-
-    It should install the rc release on Test PyPI and be able to create PDF documents from rst files
+       $ uvx -n --index-url https://test.pypi.org/simple --extra-index-url https://pypi.org/simple --prerelease allow \
+         --index-strategy unsafe-best-match rst2pdf@0.101rc1 tests/input/test_tableofcontents.rst
```
kyluca, line 97: Note this example command runs the RC version of `rst2pdf` from Test-PyPI on the `tests/input/test_tableofcontents.rst` test case. I don't have enough experience with `rst2pdf` to know if there's a better choice here, so please feel free to suggest another file.

The previous line only said it should "be able to create PDF documents from rst files", but I think supplying an example here will be more helpful.

pyproject.toml
```diff
@@ -22,14 +22,14 @@ classifiers = [
...
-    "reportlab",
-    "smartypants",
+    "docutils~=0.20.0",
+    "importlib-metadata~=8.0",
+    "jinja2~=3.0",
+    "packaging~=24.0",
+    "pygments~=2.0",
+    "pyyaml~=6.0",
+    "reportlab~=4.0",
+    "smartypants~=2.0",
```
kyluca, line 32: Placed some limitations on these versions to avoid breaking changes.

For example, "pyyaml~=6.0" means install any version in the 6.x series, but not 7.x or above which could have API breaking changes. Syntactic sugar for `"pyyaml>=6.0,<7".

pyproject.toml
```diff
@@ -87,3 +85,10 @@ exclude = '''
 packages = ["rst2pdf"]
 
 [tool.setuptools_scm]
+
+[tool.uv]
+dev-dependencies = [
+    "pymupdf~=1.0,<1.24.11",
```
kyluca, line 91: Note there's something funky going on in `pymupdf==1.24.11` as per https://github.com/pymupdf/PyMuPDF/issues/3931. From what it sounds like it's out of their control, but it does make the test results flood with warnings that we can't fix.

Opted to use lower versions to avoid the noise.

uv.lock
kyluca, line 1: Spiritual successor to the old `requirements.txt` that was generated by `pip-compile`.

kyluca:
Of course, take your time!

I'll try to figure out what's going wrong with the tests between #1241 and this PR. Something to do with the third party package updates I think because #1241 has:

```
============ 37 failed, 256 passed, 22 skipped in 319.69s (0:05:19) ============
```

But this PR #1242 has only:

```
============ 1 failed, 292 passed, 22 skipped in 287.36s (0:04:47) =============
```

The one failing test that remains is:

```
_______________________________ sphinx-issue252 ________________________________
conftest.CompareException: Text of the last printed block differs from the reference
assert ['eiπ+', '1', '=', '0'] == ['eiπ', '+', '1', '=', '0']
```

README.rst
akrabat: We removed ` pip` in favour of `pip`  due PEP 668's which impacted Linux users (#1165). 

As rst2pdf is targeted at end-users, we should provide pipx in addition to uv.

.github/workflows/ci.yml
```diff
@@ -17,28 +17,27 @@ jobs:
...
+        uvx pre-commit run --all-files --show-diff-on-failure
+      if: matrix.python-version == '3.9'
```
akrabat, line 37: The intention is to ensure that the pre-commit run only happens once as there's no benefit running it on all python versions. We should fix the matrix definitions properly or bump this to 3.12.

doc/RELEASE_PROCESS.rst
```diff
@@ -78,69 +74,47 @@ This is an outline of what needs to be done in order to release rst2pdf.
...
+       $ rm -rf build/ rst2pdf.egg-info/ dist/
 
 #. Once rc version is working, release to PyPI_ by generating official release and uploading
 
     ::
 
-       $ python -m build
-       $ twine upload --repository rst2pdf dist/*
+       $ uv build --no-sources
+       $ uv publish --token {your token here}
```
akrabat, line 110: Is there a way to store the token somewhere such that it doesn't appear in the bash history?

doc/DEVELOPERS.rst
akrabat: Good idea. It's so much faster when testing locally :)

akrabat (commented):
Comments from my first pass reading through the diff.

README.rst
kyluca: Ah that makes sense, moving away from `pip` as an installation method for system wide packages.

But now `uv tool install rst2pdf` is equivalent to `pipx install rst2pdf` as they both create and manage isolated environments internally. The only difference being `uv` is faster.

Not sure if there's any benefit to having both `pipx` and `uv` instructions here or if it would only confuse users?

.github/workflows/ci.yml
kyluca: Yeah that definitely makes sense. In the current state it was never evaluating to true and running because the matrix no longer had 3.8 (only went from 3.9 -> 3.13).

Of those two options, would you prefer me to put 3.8 into the matrix or bump the conditional to look for 3.12?

Note 3.8 is already in EOL status: https://devguide.python.org/versions/

README.rst
akrabat: I think `uv` is too new to be the only one. I can imagine developers moving to `uv` sooner than consumers.

.github/workflows/ci.yml
akrabat: 3.12, please. With 3.8 end of life, we aren't interested in supporting it any more and probably need to update our PyPI metadata on the next release too, thinking about it.

kyluca: No worries, updated this line to 3.12.

README.rst
kyluca: No worries, the original `pipx` instructions are back in now with `uv` being an additional method and not the only suggested one.

doc/RELEASE_PROCESS.rst
kyluca: I believe it can be placed in an environment variable (`UV_PUBLISH_TOKEN`), would that be better? Depending on how that's set locally for the release process, it should be an improvement on storing the old token in a plaintext config file `~/.pypirc`

https://docs.astral.sh/uv/guides/publish/#publishing-your-package

akrabat: One day, we should automate publishing to PyPI when a tag is pushed to git :)

akrabat:
Now that #1243 is merged, I think that the next steps on this one is to rebase against latest main and then set the minimum Python version to 3.9 to pickup a Sphinx >= 7.3. With any luck, that'll get the CI to pass.

akrabat:
Though, it also looks like we need to be using flake8 v6.1.0 for Python >=3.12.

pyproject.toml
```diff
@@ -87,3 +85,10 @@ exclude = '''
 packages = ["rst2pdf"]
 
 [tool.setuptools_scm]
+
+[tool.uv]
+dev-dependencies = [
```
ghickman, line 90: [uv shipped dependency-group support in 0.4.27](https://github.com/astral-sh/uv/releases/tag/0.4.27), since that's a standardised way to do groups it might make sense to use that instead of this uv-specific version?

doc/DEVELOPERS.rst
lornajane: I love both this change and the commentary that helped me understand it!

doc/DEVELOPERS.rst
```diff
@@ -50,25 +50,11 @@ activate a venv::
...
+    uv sync --all-extras
 
-Ensure that setuptools and pip are up to date::
+If you don't have ``uv``, please see `the installation docs <https://docs.astral.sh/uv/getting-started/installation/>`_
 
-    pip install --upgrade setuptools pip
-
-Now you can install rst2pdf from this source code::
-
-    pip install -c requirements.txt -e .[aafiguresupport,mathsupport,plantumlsupport,rawhtmlsupport,sphinx,svgsupport,tests]
```
lornajane, line 62: I see that we didn't replace this command with anything. So when I edit the source code, how do I get that updated version used? Run `uv sync` again?

lornajane: I think yes, I answered my own question here :)

README.rst
lornajane: Thanks! Having both uv and pipx sounds like the right current approach to me. As a user of many different tech stacks, I don't always want to be using the latest and greatest of all their dependency management tools!

doc/DEVELOPERS.rst
lornajane: I think `uv` is growing on me :)

doc/RELEASE_PROCESS.rst
```diff
@@ -78,69 +74,47 @@ This is an outline of what needs to be done in order to release rst2pdf.
...
-          username = __token__
-          password = {your token here}
-
-        [testrst2pdf]
-          repository = https://test.pypi.org/legacy/
-          username = __token__
-          password = {your token here}
-
-
-    You can get your token from the account settings section of https://pypi.org/ & https://test.pypi.org/
```
lornajane, line 105: Do we need to keep the instructions for getting the token as I think this is the one used in the new command as well?

doc/RELEASE_PROCESS.rst
lornajane: Let's leave this as it is for now - I think @akrabat is the only person to have published in recent memory, although in theory I also have access. It's fine as it is or we can adjust it whenever we next release and need these instructions.

pyproject.toml
lornajane: Thank you! I think that's a good decision.

lornajane:
This is definitely taking shape, thank you @kyluca!

I've resolved the comments that need no further action (although I must say how much I appreciated the extra context!), to try to keep our focus on what's left to do. It looks to me like we're pretty much there but we should:
 - rebase onto main to get the test fixes from #1243 
 - either adopt dependency groups or even capture it as a followup issue if we don't want to wait to put that in right now
 - bump the flake8 version

pyproject.toml
lornajane: Thanks for this tip! I adapted the structure to add a dev dependency group.

lornajane:
@akrabat could you revisit this one when you have time please? I addressed everything I could see we'd talked about in the thread.

doc/RELEASE_PROCESS.rst
akrabat: Yes we do. I've re-added.

akrabat (approved):
This is great. Thanks @lornajane. I've made a few tweaks. Please check them and if okay, we're good to merge.

lornajane (approved):
This was a pretty epic team effort, thank you @kyluca for your input - it's much appreciated and we definitely couldn't have got this far without you :) Thanks @ghickman for the pointers and @akrabat for the polishing ... I think we're all set!

PR #1243, opened 2024-11-03
Avoid adding a second PDF document with Sphinx >7.3

lornajane:
We saw in #1241  that a change in Sphinx 7.3 causes quite some weirdness and we end up with two entries in the list of `pdf_documents` in most cases. This pull request attempts to stop that from happening (unless we need that extra document config because there isn't any).

**Edit:** all tests now passing with new versions of Sphinx.


lornajane:
The `sphinx-versionmodified` test reworded the changelog entry for an addition to "Added in" rather than "New in" ... I'll update the reference file on that one.

lornajane:
This pull request has all the tests passing when Sphinx is 7.3.0 or later, which is the version expected by rst2pdf - I can't find a way to make things work across the Sphinx changes.

@akrabat I updated the reference test files for `sphinx-issue319` and `320`, these are funky because they don't have the `pdf_documents` entry in the `conf.py` file, and I think the new reality, which uses some of the provided config information, is probably fine? But see what you think - this should also unblock what @kyluca  is doing.

lornajane:
(also this builds on #1241 so when we merge this PR, that one will also show as merged.

kyluca:
> all tests now passing with new versions of Sphinx.

Nice work! 🎉

akrabat (approved):
LGTM. Thanks.

PR #1246, opened 2024-11-29
Support emphasize-lines as an alias for hl_lines

### LINKED ISSUE #1057: Support Sphinx's emphasize-lines option
Sphinx has `emphasize-lines` which seems to work similar to our `hl_lines`. It would be helpful to support Sphinx's option, presumably as an alias to `hl_lines`?

lornajane:
Fixes #1057 to add support for `emphasize-lines` (note the hyphen!) for code blocks because Sphinx does. We usually use `hl_lines` because that's how it's named in Pygments.

With this change, both versions are supported, so that using `emphasize-lines` will render with either tool.

NOTE: if syntax highlighting is broken, or the dimming doesn't work, the tests don't fail. I think the colour changes aren't severe enough. I did add a test but I'm not confident it would catch a regression on this option name. Suggestions welcome!

### DIFF
--- a/rst2pdf/directives/code_block.py
+++ b/rst2pdf/directives/code_block.py
@@ -242,7 +242,7 @@ def code_block_directive(
 
     content = content.replace('\t', ' ' * tabw)
 
-    # option hl_lines renamed to emphasize-lines for Sphinx compatibility
+    # hl_lines is the option, but if it isn't used, the alias emphasize-lines is also supported for Sphinx compatibility
     hl_lines = options.get('hl_lines', [])
     # if hl_lines isn't used, check if emphasize-lines should be
     if hl_lines == []:
--- a/rst2pdf/directives/code_block.py
+++ b/rst2pdf/directives/code_block.py
@@ -242,7 +242,12 @@ def code_block_directive(
 
     content = content.replace('\t', ' ' * tabw)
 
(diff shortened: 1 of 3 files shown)


akrabat:
Are we renaming to `emphasize-lines` and making `hl_lines` an alias? Logically we should as the Sphinx option is probably the most well-known in rst circles.

lornajane:
As I wrote it here, we're keeping `hl_lines` and supporting `emphasize-lines` as an alias. My logic is that we are wrapping/exposing Pygments features, which is our dependency. If Sphinx calls it something else then we can definitely be kind and support it, but it doesn't seem to be a docutils standard so I decided Pygments has the authority here, and that's also in line with what our existing users already expect. Definitely open to debate since I built it the other way around initially!

rst2pdf/directives/code_block.py
```diff
@@ -242,7 +242,12 @@ def code_block_directive(
 
     content = content.replace('\t', ' ' * tabw)
 
+    # option hl_lines renamed to emphasize-lines for Sphinx compatibility
```
akrabat, line 245: Let's go with `emphasize-lines` as the alias and reflect that in this comment.

akrabat (changes requested):

akrabat (approved):
LGTM

Issue #963, opened 2020-12-25
section heading appearing at bottom of page


Phillip-M-Feldman:
I periodically get a section heading appearing at the bottom of a page, with the first non-heading line of text at the top of the next page.  It would be helpful if there were a way to force headings to appear together with the text to which they apply.

lornajane:
Checking the manual, it looks like we have some limited widow/orphan support, but within paragraphs, not related to having a title separated from the thing it actually applies to. I agree this would make a good feature addition.

lornajane:
Looking for something completely different and I just found a `keepWithNext` property that you can set to true on the heading class in your style ... can you give it a go and let us know? I'm working on a manual update for the stylesheets so this would be excellent to pin down in more detail!

Phillip-M-Feldman:
Now I'm hitting the following issue: "Likely that a flowable is too large
for any frame."

lornajane:
Oh dear. That doesn't _exactly_ sound like progress, does it! Is there any danger you can turn this into a replication case? Is the thing on the next page quite tall?

Phillip-M-Feldman:
I'm assuming that a "replication case" is a small case that demonstrates
the problem.  My previous attempts to create such a case haven't been
successful.  It would be wonderful if there were a way to determine
which element of the input is causing the "flowable is too large" error.

akrabat:
"flowable is too large" occurs when something to be rendered is bigger than a page and cannot be rendered across multiple pages. The most common case is an image, but there are probably others.

akrabat:
The situation when a heading appears at the end of page occurs when the next element is not a paragraph and is something else like a table (which also covers lists of numbers/bullets). 

We even have a test for it ([test_issue_255_2.rst](https://raw.githubusercontent.com/rst2pdf/rst2pdf/9c3eee9ec94ba25280c82795465ee2736646e248/rst2pdf/tests/input/test_issue_255_2.rst)) and it is [ignored](https://github.com/rst2pdf/rst2pdf/blob/695ea6a214699cfd01272ef1b878794c7903a968/rst2pdf/tests/input/test_issue_255_2.ignore) with the explanation:

> headers have keeptogether attribute set, however if the next item is a table, then the headers do not keep with it.
> 
> Believed to be a ReportLab bug.

I imagine some investigation could take place to see if this is still true.

PR #1252, opened 2024-12-22
Reinstate a formerly skipped test, seems to work

lornajane:
I noticed that we had a test marked as skipped due to a ReportLab bug. I don't know what the bug was, but the output PDF I get now seems plausible from reading the input so I suggest we re-accept the test back into the project.


akrabat:
Unfortunately, the issue for which this test was written is still not working. The issue is that the section header doesn't end up on the next page if the element immediately following it is not a paragraph and is too big to fit on the page.

Weirdly, something's changed in our rendering since this test was written such that the table now fits on page 1. If we change the table to something that's a little larger such as:

```rst
.. raw:: pdf

   Spacer 0 21cm

H1
--

Heading
-------

+--------------------------------------------------------------------------------------------+
| This text is too long to fit onto page 1 due to the large vertical space so this table     |
| will be pushed to the next page (page 2). As the heading style is marked ``KeepWithNext``, |
| we expect that both the Heading and H1 headers will also be pushed onto page 2, leaving    |
| page 1 blank.                                                                              |
+--------------------------------------------------------------------------------------------+

```

then you can see that the table's text appears on page 2, but the headings are on page 1 still.

If you change the rst to:

```
.. raw:: pdf

   Spacer 0 21cm

H1
--

Heading
-------

This text is too long to fit onto page 1 due to the large vertical space so this paragraph
will be pushed to the next page (page 2). As the heading style is marked ``KeepWithNext``,
we expect that both the Heading and H1 headers will also be pushed onto page 2, leaving
page 1 blank.
```

Then you see that the headings are correctly on page 2 along with the paragraph of text.

akrabat:
To make this clearer for next time, I've renamed and improved the test to `test_issue_255_section_headers_keep_with_next_table.rst` and ignored it. I've also created `test_issue_255_section_headers_keep_with_next_paragraph.rst` that shows what we expect to happen.

lornajane (commented):
Thanks, this does make the problem clearer and I think we should merge this as-is. (I can't actually approve as it was my PR originally)

lornajane:
We have #963  open to handle this problem so I won't create a new issue to track it.

akrabat (approved):
Approved as per https://github.com/rst2pdf/rst2pdf/pull/1252#pullrequestreview-2530960898

PR #708, opened 2018-11-10
Removal of smartypants

ralsina:
Removed our copy of Smartypants and added it as a dependency. Fix #694

PR #1253, opened 2024-12-22
Remove smartypants, use the docutils version

### LINKED ISSUE #1244: SmartyPants doesn't support 3.12
**Description of problem**

[Smartypants 2.0.1](https://github.com/leohemsted/smartypants.py) does not support Python 3.12, so we need to decide what to do about this.

It was [last released to PyPI](https://pypi.org/project/smartypants/#history) in 2017.

lornajane:
Fixes #1244 where we depend on smartypants but it's no longer actively maintained and has problems with newer versions of Python.

Looks like Sphinx or docutils ran into this and wrote the replacement function we need (I love open source) but something funky is going on - the same configuration options are supported as in our manual, but the output isn't quite right and two of the tests fail so I need to take another look before we can properly review//merge.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,6 @@ dependencies = [
     "pygments~=2.0",
     "pyyaml~=6.0",
     "reportlab~=4.0",
-    "smartypants~=2.0",
 ]
 description = "Convert reStructured Text to PDF via ReportLab."
 dynamic = ["version"]
--- a/rst2pdf/basenodehandler.py
+++ b/rst2pdf/basenodehandler.py
@@ -35,8 +35,8 @@ from copy import copy
 import inspect
 import types
 
-from smartypants import smartypants
 import docutils.nodes
+from docutils.utils import smartquotes
 
 from .flowables import BoundByWidth, TocEntry
 from .log import log, nodeid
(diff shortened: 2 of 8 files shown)


akrabat:
This is quite interesting as looking at https://github.com/rst2pdf/rst2pdf/pull/708, I think we used to have our own implementation of smartypants that worked the same way as the docutils one did and then we replaced it with [leohemsted/smartypants.py(https://github.com/leohemsted/smartypants.py) which behaves differently.

I say this because the [reference PDF](https://github.com/rst2pdf/rst2pdf/blob/a0d45c800aff2f1861e282c6722ceb497fef975f/rst2pdf/tests/reference/test_smarty.pdf) we deleted in #708 has the same sort of quotes for the last line as the docutils output in this PR does.

lornajane:
Oh that is interesting! I am not sure it's correct though, the source has escaped quote characters.

akrabat:
I could be wrong, but in the Sphinx reference PDF, the `\`\`A B'' has curly quotes around A and B` line has rendered the opening quotes as single curly double quotes character, but the closing is two single curly quote characters.

In the non-Sphinx version both the opening and closing are a single curly double quote character.

akrabat:
Also, while we're here, should we add a test that that a single `'` is converted to a curly quote or do we have that in another test?

lornajane:
The single quote is a good shout, I don't think we have that elsewhere! I'm tempted to add a test that uses the same source but different config options as well ....
 I'll check the Sphinx output

akrabat (approved):
LGTM. Thanks!

PR #1272, opened 2025-05-10
drop usage of old importlib_metadata

a-detiste:
I think that importlib_metadata could be removed from Debian before end of 2025, let's see how it fares.


a-detiste:
@stephenfin

PR #1301, opened 2026-03-09
Update release process to use this version of rst2pdf

akrabat:
This version of rst2pdf is the correct version to use as we know that the dependencies are installed.


lornajane (approved):
makes sense. Although I think I had problems with the version because I had untracked and/or uncommitted files in my local tree. I think I ended up doing part of it from a fresh clone but this is a good addition and we can release more often to harden the instructions!

